PR #97, opened 2024-11-19
fix: set max dimension to 1600 in readme

usamoi:


VoVAllen:
Please update the code also https://github.com/tensorchord/VectorChord/blob/d0d84616e04bfb0379f183cd15d0c54de04fe1c4/src/index/am.rs#L232

PR #106, opened 2024-11-21
fix: optimize insertions in building when lists = 1

usamoi:


kemingy:
If you change the toolchain version, you may need to update the image. I'll create a CI to push the image to GitHub packages so developers can trigger the CI to update the image.

usamoi:
> If you change the toolchain version, you may need to update the image. I'll create a CI to push the image to GitHub packages so developers can trigger the CI to update the image.

Why do we need to update the image? Why not add targets when image is used?

kemingy:
> > If you change the toolchain version, you may need to update the image. I'll create a CI to push the image to GitHub packages so developers can trigger the CI to update the image.
> 
> Why do we need to update the image? Why not add targets when image is used?

Adding targets later means you need to do it for every docker image run. Updating the version is not a highly frequent operation.

usamoi:
> > > If you change the toolchain version, you may need to update the image. I'll create a CI to push the image to GitHub packages so developers can trigger the CI to update the image.
> > 
> > 
> > Why do we need to update the image? Why not add targets when image is used?
> 
> Adding targets later means you need to do it for every docker image run. Updating the version is not a highly frequent operation.

I think adding a target takes less than one second on Github actions and downloading the docker image takes more than one. If we want it to be faster, we should remove docker for first.

usamoi:
I just noticed that sqllogictest tests does not run on all PostgreSQL major versions.

kemingy:
Adding targets costs about 17.8s in GitHub CI. Downloading the whole image is about 53s for now, almost the same as downloading from Docker Hub. ~~I thought accessing ghcr should be much faster.~~

usamoi:
> Adding targets costs about 17.8s in GitHub CI.

Screenshot_20241121_204825

Their servers must be on the moon or Mars. It takes only 4 seconds even on my machine.

PR #141, opened 2024-12-17
feat: indexing on scalar8

usamoi:
indexing on scalar8 directly brings loss of accuracy (0.95 -> 0.77 on glove-100) and latency grows up to 500 ms


usamoi:
@VoVAllen

VoVAllen:
Why? I think this is not consistent with your previous experiement?

VoVAllen:
It can be slower but 500ms is way too slow. And the recall drop too much still doesn't make sende to me

usamoi:
QPS 44 Recall 0.77

PR #149, opened 2024-12-23
chore: type check and external test

### LINKED ISSUE #144: feat: external centroid vector type check

cutecutecat:
Close #144

### DIFF
--- a/src/vchordrq/algorithm/build.rs
+++ b/src/vchordrq/algorithm/build.rs
@@ -203,14 +203,34 @@ impl Structure {
     ) -> Vec<Self> {
         use std::collections::BTreeMap;
         let VchordrqExternalBuildOptions { table } = external_build;
-        let query = format!("SELECT id, parent, vector FROM {table};");
+        let dump_query = format!("SELECT id, parent, vector FROM {table};");
+        let table_name = table.split('.').last().unwrap().to_string();
+        let type_check_query = format!(
+            "SELECT COUNT(*)::INTEGER 
+            FROM pg_catalog.pg_extension e
+            LEFT JOIN pg_catalog.pg_namespace n ON n.oid = e.extnamespace
+            LEFT JOIN information_schema.columns i ON i.udt_schema = n.nspname
+            WHERE e.extname = 'vector' AND i.udt_name = 'vector' 
+            AND i.table_name = '{table_name}' AND i.column_name = 'vector';"
+        );
         let mut parents = BTreeMap::new();
         let mut vectors = BTreeMap::new();
         pgrx::spi::Spi::connect(|client| {
             use crate::datatype::memory_pgvector_vector::PgvectorVectorOutput;
             use base::vector::VectorBorrowed;
             use pgrx::pg_sys::panic::ErrorReportable;
-            let table = client.select(&query, None, None).unwrap_or_report();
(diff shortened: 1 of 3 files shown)


src/vchordrq/algorithm/build.rs
```diff
@@ -220,7 +240,7 @@ impl Structure {
                 let pop = parents.insert(id, parent);
                 if pop.is_some() {
                     pgrx::error!(
-                        "external build: there are at least two lines have same id, id = {id}"
+                        "extern build: there are at least two lines have same id, id = {id}"
```
usamoi, line 243: Screenshot_20241224_133921

src/vchordrq/algorithm/build.rs
```diff
@@ -203,14 +203,34 @@ impl Structure {
     ) -> Vec<Self> {
         use std::collections::BTreeMap;
         let VchordrqExternalBuildOptions { table } = external_build;
-        let query = format!("SELECT id, parent, vector FROM {table};");
+        let dump_query = format!("SELECT id, parent, vector FROM {table};");
+        let table_name = table.split('.').last().unwrap().to_string();
+        let type_check_query = format!(
```
usamoi, line 208: We'd better get schema name of extension vector here by querying `pg_extension` and `pg_namespace`. So that we could get oid of `vector` by `regtypein('SCHEMA_NAME.vector')` (and `halfvec`).

If we want to support `halfvec`, we could then query

```
SELECT atttypid FROM pg_attribute WHERE attrelid = '{table}'::regclass AND attname = 'vector';
```

to know what type it exactly is.

If we don't want to support `halfvec`, we could just query `SELECT id, parent, vector::{vector_schema}.vector FROM {table};` for centroids.

cutecutecat: OK, I prefer to not support halfvec and force cast it.

src/vchordrq/algorithm/build.rs
```diff
@@ -203,28 +203,54 @@ impl Structure {
...
             use pgrx::pg_sys::panic::ErrorReportable;
-            let table = client.select(&query, None, None).unwrap_or_report();
-            for row in table {
+            // Get the schema of pgvector
+            let schema_query = format!(
+                "SELECT n.nspname::TEXT 
+                FROM pg_catalog.pg_extension e
+                LEFT JOIN pg_catalog.pg_namespace n ON n.oid = e.extnamespace
+                LEFT JOIN information_schema.columns i ON i.udt_schema = n.nspname
+                WHERE e.extname = 'vector' AND i.table_name = '{table_name}' AND i.column_name = 'vector';");
```
usamoi, line 219: Don't reference into system views

No need to reference the table and the column

src/vchordrq/algorithm/build.rs
```diff
@@ -203,28 +203,54 @@ impl Structure {
...
+                })
+                .collect();
+            // Check the `vector` column is pgvector-based type
+            let pgvector_schema = if let [schema] = &nspname[..] {
+                schema.clone()
+            } else {
+                pgrx::error!(
+                    "external build: `vector` column should be a pgvector type at the external table"
+                );
+            };
```
usamoi, line 236: Don't; you shouldn't.

usamoi (approved):

PR #176, opened 2025-01-26
fix: reserve capacity

usamoi:


usamoi:
Why CI does not run?

VoVAllen:
https://github.com/tensorchord/VectorChord/blob/main/.github/workflows/rust.yml#L4-L11

usamoi:
Tests shows it's a debuff.

VoVAllen:
I don't understand why reserve capacity for vec will degrade the performance

usamoi:
> I don't understand why reserve capacity for vec will degrade the performance

To know exact capacity, the process save it in a vector when scanning and make heap on it. The vector becomes bigger since the elements become bigger. It's unfriendly to cache.

VoVAllen:
We don't need exact size I think. Just some estimation should work?

VoVAllen:
Our target is just to reduce the number of resizing not really eliminate it

usamoi:
replaced by #182

PR #196, opened 2025-02-20
refactor: split quantized data vectors to two tapes

usamoi:
split data vectors to two tapes, one is immutable since `maintain`, and another one is appendable for newly-inserted vectors


usamoi:
> what's this

Changes needed by multi-vector index.

VoVAllen:
please add more detailed descriptions in the pr

PR #219, opened 2025-03-24
feat: emit INFO when performing kmeans

usamoi:
1. add build phase names
2. emit kmeans log
3. emit errors if both `Dot` and `residual_quantiztion` are used
4. update `tuples_done` by estimated value before insertions
5. update `tuples_done` after insertions are done
6. emit errors if `vchordrq_prewarm` is given a partition index
7. rename `maxsim_ip` to `maxsim`

### DIFF
--- a/crates/algorithm/src/types.rs
+++ b/crates/algorithm/src/types.rs
@@ -51,7 +51,7 @@ pub enum VectorKind {
 #[serde(deny_unknown_fields)]
 #[validate(schema(function = "Self::validate_self"))]
 pub struct VectorOptions {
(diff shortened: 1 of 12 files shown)


VoVAllen:
> emit errors if both Dot and residual_quantiztion are used
WARNING should be fine?

> emit errors if vchordrq_prewarm is given a partition index
Why? How should prewarm work with partition table and index?

usamoi:
> Why? How should prewarm work with partition table and index?

It never works. This pr improves error message.

It could work as long as prewarm function prewarms all partitions.

PR #244, opened 2025-05-07
feat: prefetch in searching

usamoi:
Adds a GUC `io_search`, with the same possible values of `io_rerank`.

notes:

* prefetching for the bottom-level bit vectors is implemented.
* prefetching for the other levels of bit vectors is not implemented yet.


VoVAllen:
Can you show the number of pure cold query comparing to the 0.3?

VoVAllen:
> Adds a GUC io_search, with the same possible values of io_rerank.

Why not merge them into one GUC?

usamoi:
> > Adds a GUC io_search, with the same possible values of io_rerank.
> 
> Why not merge them into one GUC?

This option is only intended for debugging. It's common for memory to accommodate a bit vector but not a full vector so we may control these two behaviors separately.

usamoi:
> Then we might add debug prefix to it, like debug_direct_io in pg17

If a user's environment requires modifying this GUC, using a GUC without the "debug" looks like a normal process, not a workaround for "cleaning up after bad software".

PR #245, opened 2025-05-07
feat: prefilter index

cutecutecat:
## Feat
- Allow non-heap prefilter: enabled when `rerank_in_table=false(default)` and `SET vchordrq.prererank_filtering=on`
- Support streamio in heap prefilter and heap postfilter: enabled when `rerank_in_table=true` and `SET vchordrq.io_rerank = read_stream(default)`

## Optimization
- Eliminating computational costs at heap prefilter and heap postfilter(when `rerank_in_table=true`)

### DIFF
--- a/crates/algorithm/src/rerank.rs
+++ b/crates/algorithm/src/rerank.rs
@@ -27,26 +27,30 @@ pub fn how(index: impl RelationRead) -> RerankMethod {
     }
 }
 
-pub struct Reranker<T, F, P> {
+pub struct Reranker<T, F, P, O> {
     prefetcher: P,
     cache: BinaryHeap<Result>,
     f: F,
-    _phantom: PhantomData<fn(T) -> T>,
+    _phantom: PhantomData<(fn(T) -> T, O)>,
 }
 
-impl<'b, T, F, P> Iterator for Reranker<T, F, P>
+impl<'b, T, F, P, O: Operator> Iterator for Reranker<T, F, P, O>
 where
-    F: FnMut(NonZero<u64>, Vec<<P::R as RelationRead>::ReadGuard<'_>>, u16) -> Option<Distance>,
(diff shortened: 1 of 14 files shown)


VoVAllen:
I think we should just name prererank_filtering as prefilter

crates/algorithm/src/rerank.rs
```diff
@@ -69,6 +69,7 @@ pub fn rerank_index<
 >(
     vector: O::Vector,
     prefetcher: P,
+    mut prefilter: impl FnMut(NonZero<u64>) -> bool,
```
usamoi, line 72: don't do filtering here; do filtering in `scanners`

src/index/am/mod.rs
```diff
@@ -600,6 +600,43 @@ impl SearchFetcher for HeapFetcher {
             Some((&self.values, &self.is_nulls))
         }
     }
+
+    fn filter(&mut self, key: [u16; 3]) -> bool {
```
usamoi, line 604: don't fetch the same tuple twice due to filtering and fetching the vector

src/index/scanners/default.rs
```diff
@@ -408,3 +324,81 @@ impl SearchBuilder for DefaultBuilder {
...
+                rerank_index::<O, _, _>(vector, prefetcher, prefilter)
+                    .map(move |(distance, payload)| (opfamily.output(distance), payload)),
+            )
+        }
+    }
+}
+
+#[inline(always)]
+fn rerank_heap_wrapper<'a, O: Operator>(
+    vector: O::Vector,
```
usamoi, line 374: don't handle `rerank_heap` and `rerank_index` differentl; split `algorithm::insert` into two functions, `algorithm::insert_vector` and `algorithm::insert_index`, so that the full-precision vectors in the index are unprojected, too

usamoi (changes requested):

PR #246, opened 2025-05-07
add vectorchord docs link

xieydd:


mertalev:
Is this true for <1M vectors as well? I remember seeing a mention that it should be `vectors / 1000` below 1M. But going from 900 lists at 900k to 4096 at 1M is a very big jump, so one of the two `lists` has to be suboptimal.

Anecdotally, I'm getting reasonable results with:
<128k: 1 list
>=128k, <2M: `2 * vectors / 1000`
>=2M: `4 * sqrt(vectors)`

xieydd:
> Is this true for <1M vectors as well? I remember seeing a mention that it should be `vectors / 1000` below 1M. But going from 900 lists at 900k to 4000 at 1M is a very big jump, so one of the two `lists` has to be suboptimal.
> 
> Anecdotally, I'm getting reasonable results with:
> 
> ```
> <128k: 1 list
> >=128k, <2M: 2 * vectors / 1000
> >=2M: 4 * sqrt(vectors)
> ```

Thanks for your feedback @mertalev , can you confirm that @cutecutecat ?

cutecutecat:
> > Is this true for <1M vectors as well? I remember seeing a mention that it should be `vectors / 1000` below 1M. But going from 900 lists at 900k to 4000 at 1M is a very big jump, so one of the two `lists` has to be suboptimal.
> > Anecdotally, I'm getting reasonable results with:
> > ```
> > <128k: 1 list
> > >=128k, <2M: 2 * vectors / 1000
> > >=2M: 4 * sqrt(vectors)
> > ```
> 
> Thanks for your feedback @mertalev , can you confirm that @cutecutecat ?

Yes, 1M is an estimated upper bound. We haven't carefully do experiments for this argument when the vectors size is little. 

I think @mertalev 's result could be more precise.

README.md
```diff
@@ -101,6 +101,16 @@ lists = []
 $$);
 ```
 
+> [!NOTE]
+> The `lists` option should be configured based on the number of vectors. Below is a table to assist with your selection.
+
+| vectors Range | List Calculation Formula       | Example Result   |
+| ------------- | ------------------------------ | ---------------- |
+| <128k         | list = 1                       | 1                |
+| ≥128k and <2M | list = (2 * vectors) / 1000    | [256, 4000]      |
```
VoVAllen, line 110: `256 - 4000` looks better. User may think `[256, 4000]` as a python list instead of a closed interval

cutecutecat (approved):

README.md
xieydd: I will explain below

Issue #243, opened 2025-05-06
docker run failed!


xiaojunxiang2023:
[root@~]# docker run   --name vectorchord-demo   -e POSTGRES_PASSWORD=mysecretpassword   -p 5438:5432   -d ghcr.io/tensorchord/vchord-postgres:pg17-v0.3.0
7444c698ba759328928a6828463bab548c0b5ebc3f6b5abf34a1ae2c01af8708

[root@~]# docker logs vectorchord-demo 
**could not execute command ""/usr/lib/postgresql/17/bin/postgres" -V": Cannot allocate memory**
initdb: error: program "postgres" is needed by initdb but was not found in the same directory as "/usr/lib/postgresql/17/bin/initdb"

[root@~]# free -mh
                    total        used        free      shared  buff/cache   available
Mem:           251G         41G         20G         12G        190G        196G
Swap:            0B          0B          0B

xiaojunxiang2023:
> It seems to be a problem with insufficient memory. Can you try a smaller shared_buffer setting?

"My operation(docker run...) is based on the instructions at https://github.com/tensorchord/VectorChord?tab=readme-ov-file#quick-start. If I need to adjust parameters, how should I do it?"

xiaojunxiang2023:
> It seems to be a problem with insufficient memory. Can you try a smaller shared_buffer setting?

Even though I specified --shm-size=16g, the error message hasn't changed. My machine information is shown in the image.
![Image](https://github.com/user-attachments/assets/7d1fd0c6-a545-483c-813e-d1b728a43402)

![Image](https://github.com/user-attachments/assets/4ef73084-5400-455f-beb9-2cdd759c6e98)

usamoi:
Looks like https://github.com/docker-library/postgres/issues/1100#issuecomment-1599660628.

Can you update Docker and `libseccomp` on your host machine?

xiaojunxiang2023:
> Looks like [docker-library/postgres#1100 (comment)](https://github.com/docker-library/postgres/issues/1100#issuecomment-1599660628).
> 
> Can you update Docker and `libseccomp` on your host machine?

I was using Docker 20.x before, and after switching to Docker 27.x, the environment started up successfully without any issues.
It’s likely due to the Docker version.

xiaojunxiang2023:
> Looks like [docker-library/postgres#1100 (comment)](https://github.com/docker-library/postgres/issues/1100#issuecomment-1599660628).
> 
[...]

PR #250, opened 2025-05-13
fix docker run failed 'Cannot allocate memory'

xiaojunxiang2023:
https://github.com/tensorchord/VectorChord/issues/243


usamoi:
The root cause is that your environment is too outdated. The `--privileged` option poses security risks and shoud not be used by most users.

PR #284, opened 2025-07-10
feat: query recall eval func

cutecutecat:
## Usages

```sql
> SELECT vchordrq_evaluate_query_recall(query_text =>$$SELECT ctid FROM test ORDER BY val <-> '[0.5, 0.25, 1.0]' LIMIT 10$$);
 vchordrq_evaluate_query_recall 
--------------------------------
                              1
(1 row)

> SELECT vchordrq_evaluate_query_recall(query_text => $$SELECT ctid FROM test ORDER BY val <-> '[0.5, 0.25, 1.0]' LIMIT 100$$, exact_search=>true);
 vchordrq_evaluate_query_recall 
--------------------------------
                              1

> SELECT vchordrq_evaluate_query_recall(query_text => $$SELECT ctid FROM test WHERE FALSE' LIMIT 100$$, exact_search=>true);
 vchordrq_evaluate_query_recall 
--------------------------------
                              NaN
(1 row)
```

### DIFF
--- a/src/index/gucs.rs
+++ b/src/index/gucs.rs
@@ -36,6 +36,8 @@ pub enum PostgresIo {
     ReadStream,
 }
 
+static VCHORDG_ENABLE_SCAN: GucSetting<bool> = GucSetting::<bool>::new(true);
+
 static VCHORDG_EF_SEARCH: GucSetting<i32> = GucSetting::<i32>::new(64);
 
 static VCHORDG_BEAM_SEARCH: GucSetting<i32> = GucSetting::<i32>::new(1);
@@ -55,6 +57,8 @@ static VCHORDG_IO_RERANK: GucSetting<PostgresIo> = GucSetting::<PostgresIo>::new
     #[cfg(any(feature = "pg17", feature = "pg18"))]
     PostgresIo::ReadStream,
 );
+
+static VCHORDRQ_ENABLE_SCAN: GucSetting<bool> = GucSetting::<bool>::new(true);
 
 static VCHORDRQ_PROBES: GucSetting<Option<CString>> = GucSetting::<Option<CString>>::new(Some(c""));
 
@@ -83,6 +87,14 @@ static VCHORDRQ_IO_RERANK: GucSetting<PostgresIo> = GucSetting::<PostgresIo>::ne
 );
 
 pub fn init() {
+    GucRegistry::define_bool_guc(
+        c"vchordrq.enable_scan",
+        c"`enable_scan` argument of vchordrq.",
+        c"`enable_scan` argument of vchordrq.",
+        &VCHORDRQ_ENABLE_SCAN,
+        GucContext::Userset,
+        GucFlags::default(),
+    );
     GucRegistry::define_string_guc(
         c"vchordrq.probes",
         c"`probes` argument of vchordrq.",
@@ -161,6 +173,14 @@ pub fn init() {
         #[cfg(any(feature = "pg15", feature = "pg16", feature = "pg17", feature = "pg18"))]
         pgrx::pg_sys::MarkGUCPrefixReserved(c"vchordrq".as_ptr());
     }
+    GucRegistry::define_bool_guc(
+        c"vchordg.enable_scan",
+        c"`enable_scan` argument of vchordg.",
+        c"`enable_scan` argument of vchordg.",
+        &VCHORDG_ENABLE_SCAN,
+        GucContext::Userset,
+        GucFlags::default(),
+    );
     GucRegistry::define_int_guc(
         c"vchordg.ef_search",
         c"`ef_search` argument of vchordg.",
@@ -215,6 +235,10 @@ pub fn init() {
     }
 }
 
+pub fn vchordg_enable_scan() -> bool {
+    VCHORDG_ENABLE_SCAN.get()
+}
+
 pub fn vchordg_ef_search() -> u32 {
     VCHORDG_EF_SEARCH.get() as u32
 }
@@ -246,6 +270,10 @@ pub fn vchordg_io_rerank() -> crate::index::vchordg::scanners::Io {
         #[cfg(any(feature = "pg17", feature = "pg18"))]
         PostgresIo::ReadStream => Io::Stream,
     }
+}
+
+pub fn vchordrq_enable_scan() -> bool {
+    VCHORDRQ_ENABLE_SCAN.get()
 }
 
 pub fn vchordrq_probes() -> Vec<u32> {
--- a/src/sql/finalize.sql
+++ b/src/sql/finalize.sql
@@ -148,6 +148,75 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordrq_amhan
 CREATE FUNCTION vchordrq_prewarm(regclass, integer default 0) RETURNS TEXT
 STRICT LANGUAGE c AS 'MODULE_PATHNAME', '_vchordrq_prewarm_wrapper';
 
+CREATE FUNCTION vchordrq_evaluate_query_recall(
+    query text,
+    exact_search boolean default false,
+    accu_epsilon real default 1.9
+)
+RETURNS real
+LANGUAGE plpgsql
+STRICT
+AS $$
+DECLARE
+    rough tid[];
+    accu tid[];
+    match_count integer := 0;
+    accu_k integer;
+    recall real;
+    rough_probes text;
+    accu_probes text;
+BEGIN
+    IF query LIKE '%@#%' AND NOT exact_search THEN
+        RAISE EXCEPTION 'MaxSim operator cannot be used for estimated recall evaluation. Please use exact_search => true.';
+    END IF;
+    IF NOT exact_search THEN
+        BEGIN
+            rough_probes := current_setting('vchordrq.probes');
+        END;
+    END IF;
+
+    BEGIN
+        EXECUTE
+            format('SELECT coalesce(array_agg(id), array[]::tid[]) FROM (%s) AS result(id)', query)
+        INTO
+            rough;
+    EXCEPTION WHEN OTHERS THEN
+        RAISE EXCEPTION 'Error executing ANN query "%": %', query, SQLERRM;
+    END;
+
+    BEGIN
+        IF exact_search THEN
+            SET LOCAL vchordrq.enable_scan = off;
+        ELSE
+            IF rough_probes = '' THEN
+                accu_probes := '';
+            ELSIF position(',' in rough_probes) > 0 THEN
+                accu_probes := '65535,65535';
+            ELSE
+                accu_probes := '65535';
+            END IF;
+            EXECUTE format('SET LOCAL "vchordrq.probes" = %L', accu_probes);
+            EXECUTE format('SET LOCAL "vchordrq.epsilon" = %L', accu_epsilon);
+            SET LOCAL vchordrq.max_scan_tuples = -1;
+        END IF;
+        EXECUTE
+            format('SELECT coalesce(array_agg(id), array[]::tid[]) FROM (%s) AS result(id)', query)
+        INTO
+            accu;
+    EXCEPTION WHEN OTHERS THEN
+         RAISE EXCEPTION 'Error executing Ground Truth query "%": %', query, SQLERRM;
+    END;
+    accu_k := cardinality(accu);
+    IF accu_k = 0 THEN
+        RAISE WARNING  'Query "%": No results found, returning NaN for recall.', query;
+        RETURN 'NaN';
+    END IF;
+    SELECT COUNT(*) INTO match_count FROM (SELECT unnest(rough) INTERSECT SELECT unnest(accu)) AS tids;
+    recall := match_count::real / accu_k::real;
+    RETURN recall;
+END;
+$$;
+
 CREATE FUNCTION vchordg_amhandler(internal) RETURNS index_am_handler
 IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhandler_wrapper';
 
--- a/tests/vchordrq/recall.slt
+++ b/tests/vchordrq/recall.slt
@@ -1,0 +1,53 @@
+statement ok
+CREATE TABLE t (val vector(3));
+
+statement ok
+INSERT INTO t (val)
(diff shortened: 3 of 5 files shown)


src/sql/finalize.sql
```diff
@@ -154,6 +154,111 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+    query_epsilon text;
+    gt_probes text;
+    enable_indexscan_original text;
+    enable_bitmapscan_original text;
+    enable_seqscan_original text;
+
+BEGIN
+    enable_indexscan_original := current_setting('enable_indexscan');
+    enable_bitmapscan_original := current_setting('enable_bitmapscan');
+    enable_seqscan_original := current_setting('enable_seqscan');
```
usamoi: These GUCs affect all indexes, not just the vector indexes in the query. Add GUCs to control the vector index specifically.

src/sql/finalize.sql
```diff
@@ -154,6 +154,111 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+    query_text text,
+    exact_search boolean default False,
+    gt_epsilon real default 1.9
+)
+RETURNS real
+LANGUAGE plpgsql
+STRICT
+AS $$
+DECLARE
+    query_results vector[];
```
usamoi: Why it's `vector[]`?

src/sql/finalize.sql
```diff
@@ -154,6 +154,111 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+BEGIN
+    enable_indexscan_original := current_setting('enable_indexscan');
+    enable_bitmapscan_original := current_setting('enable_bitmapscan');
+    enable_seqscan_original := current_setting('enable_seqscan');
+    IF NOT exact_search THEN
+        BEGIN
+            query_probes := current_setting('vchordrq.probes');
+            query_epsilon := current_setting('vchordrq.epsilon');
+        EXCEPTION WHEN OTHERS THEN
+            RAISE EXCEPTION 'Could not read vchordrq GUCs. Is the vchordrq extension loaded and configured? Error: %', SQLERRM;
```
```suggestion
            RAISE EXCEPTION 'Could not read vchordrq GUCs. Is vchord shared library loaded? Error: %', SQLERRM;
```

src/sql/finalize.sql
```diff
@@ -154,6 +154,111 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+    EXECUTE format('SET LOCAL enable_indexscan = %L', enable_indexscan_original);
+    EXECUTE format('SET LOCAL enable_bitmapscan = %L', enable_bitmapscan_original);
+    EXECUTE format('SET LOCAL enable_seqscan = %L', enable_seqscan_original);
+
+    gt_count := array_length(gt_results, 1);
+    IF gt_count IS NULL OR gt_count = 0 THEN
+        RAISE WARNING 'Ground Truth query returned 0 results. Please check and run the query_text.';
+        RETURN 0.0;
+    END IF;
+    IF gt_count != top_k THEN
```
usamoi: Shouldn't we run groundtruth query first to get `top_k`?

tests/vchordrq/recall.slt
```diff
@@ -0,0 +1,54 @@
...
+CREATE INDEX ON t USING vchordrq (val vector_l2_ops);
+
+statement ok
+SET vchordrq.epsilon = 0.8;
+
+statement ok
+SET vchordrq.probes = '1';
+
+statement error Error executing ANN query
+SELECT * from vchordrq_evaluate_query_recall(query_text =>$$SELECT val FROM t ORDER BY val <-> '[0.5, 0.25, 1.0]' LIMIT 10$$);
```
usamoi: Let's use `ctid` here, and checks the type.

src/sql/finalize.sql
```diff
@@ -154,6 +154,111 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+        SET LOCAL enable_seqscan = on;
+    ELSE
+        IF query_probes = '' THEN
+            gt_probes := '';
+        ELSIF position(',' in query_probes) > 0 THEN
+            gt_probes := '9999,9999';
+        ELSE
+            gt_probes := '9999';
+        END IF;
+        EXECUTE format('SET LOCAL "vchordrq.probes" = %L', gt_probes);
```
usamoi: We may start a subtransaction here.

cutecutecat: PL/pgSQL doesn't support a transaction inside.

https://stackoverflow.com/questions/14550241/postgresql-cannot-begin-end-transactions-in-pl-pgsql

src/sql/finalize.sql
cutecutecat: If the groundtruth first, then we should set `vchordrq.probes` and `vchordrq.epsilon` for both gt and ann query.

If ann first, it can shared the local environment GUC settings.

src/index/gucs.rs
```diff
@@ -83,6 +87,14 @@ static VCHORDRQ_IO_RERANK: GucSetting<PostgresIo> = GucSetting::<PostgresIo>::ne
...
 
 pub fn init() {
+    GucRegistry::define_bool_guc(
+        c"vchordrq.enable",
+        c"`enable` argument of vchordrq.",
+        c"`enable` argument of vchordrq.",
+        &VCHORDRQ_ENABLE,
+        GucContext::Userset,
+        GucFlags::default(),
+    );
```
usamoi: Use a longer name `vchordrq.enable_scan`.

src/sql/finalize.sql
```diff
@@ -154,6 +154,87 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+        ELSE
+            IF query_probes = '' THEN
+                gt_probes := '';
+            ELSIF position(',' in query_probes) > 0 THEN
+                gt_probes := '9999,9999';
+            ELSE
+                gt_probes := '9999';
+            END IF;
+            EXECUTE format('SET LOCAL "vchordrq.probes" = %L', gt_probes);
+            EXECUTE format('SET LOCAL "vchordrq.epsilon" = %L', gt_epsilon);
```
usamoi: What if `vchordrq.max_scan_tuples`, `vchordrq.maxsim_refine`, `vchordrq.maxsim_threshold` are used by users?

I still suspect that we shouldn't be using the index as ground truth at all.

src/sql/finalize.sql
```diff
@@ -154,6 +154,87 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+    query_epsilon text;
+    gt_probes text;
+
+BEGIN
+    IF NOT exact_search THEN
+        BEGIN
+            query_probes := current_setting('vchordrq.probes');
+            query_epsilon := current_setting('vchordrq.epsilon');
+        EXCEPTION WHEN OTHERS THEN
+            RAISE EXCEPTION 'Could not read vchordrq GUCs. Is vchord shared library loaded? Error: %', SQLERRM;
```
usamoi: I think this part is a bit redundant. It's unlikely to happen.

src/sql/finalize.sql
```diff
@@ -154,6 +154,87 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+        END;
+    END IF;
+
+    BEGIN
+        FOR iter_ctid IN EXECUTE query_text LOOP
+            query_results := array_append(query_results, iter_ctid);
+        END LOOP;
```
```sql
EXECUTE
    format('SELECT array_agg(id) FROM (%s) AS result(id)', query)
INTO
    rough;
```

src/sql/finalize.sql
```diff
@@ -154,6 +154,87 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+    EXCEPTION WHEN OTHERS THEN
+        RAISE EXCEPTION 'Error executing ANN query "%": %', query_text, SQLERRM;
+    END;
+
+    top_k := array_length(query_results, 1);
+    IF top_k IS NULL OR top_k = 0 THEN
+        RAISE WARNING 'ANN query returned 0 results. Please check and run the query_text.';
+        RETURN 0.0;
+    END IF;
```
usamoi: Don't check it here. Do the second query for first.

src/sql/finalize.sql
```diff
@@ -154,6 +154,87 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+            EXECUTE format('SET LOCAL "vchordrq.epsilon" = %L', gt_epsilon);
+        END IF;
+        FOR iter_ctid IN EXECUTE query_text LOOP
+            gt_results := array_append(gt_results, iter_ctid);
+        END LOOP;
+    EXCEPTION WHEN OTHERS THEN
+         RAISE EXCEPTION 'Error executing Ground Truth query "%": %', query_text, SQLERRM;
+    END;
+
+    gt_count := array_length(gt_results, 1);
```
usamoi: `array_length` never returns 0. Use `cardinality`.

src/sql/finalize.sql
```diff
@@ -154,6 +154,87 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+                gt_probes := '9999,9999';
+            ELSE
+                gt_probes := '9999';
+            END IF;
+            EXECUTE format('SET LOCAL "vchordrq.probes" = %L', gt_probes);
+            EXECUTE format('SET LOCAL "vchordrq.epsilon" = %L', gt_epsilon);
+        END IF;
+        FOR iter_ctid IN EXECUTE query_text LOOP
+            gt_results := array_append(gt_results, iter_ctid);
+        END LOOP;
```
```
EXECUTE
    format('SELECT array_agg(id) FROM (%s) AS result(id)', query)
INTO
    accu;
```

src/sql/finalize.sql
```diff
@@ -154,6 +154,87 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordg_amhand
...
+         RAISE EXCEPTION 'Error executing Ground Truth query "%": %', query_text, SQLERRM;
+    END;
+
+    gt_count := array_length(gt_results, 1);
+    IF gt_count IS NULL OR gt_count = 0 THEN
+        RAISE WARNING 'Ground Truth query returned 0 results. Please check and run the query_text.';
+        RETURN 0.0;
+    END IF;
+    IF gt_count != top_k THEN
+         RAISE NOTICE 'Ground Truth query returned % results, while ANN query returned % results.', gt_count, top_k;
```
usamoi: If the result is 0 or NaN, the user will always notice it. There's no need to check.

VoVAllen:
Let's say we have record the vectors in another table/view. How would the syntax look like? Shall we introduce template-like syntax  like `ORDER BY val <-> $1` or it can be done with some simple string concat?

cutecutecat:
> Let's say we have recorded the vectors in another table/view. How would the syntax look like? Shall we introduce template-like syntax like ORDER BY val <-> $1 or it can be done with some simple string concat?

I tend to implement the concated SQL query also in that provided table like:

```shell
 table_name | column_name | operator |     vector_text      | simplified_query 
------------+-------------+----------+----------------------+------------------
 test       | val         | <->      | '[0.50, 0.25, 1.00]' | SELECT ctid FROM test ORDER BY val <-> '[0.5, 0.25, 1.0]';
```

src/index/gucs.rs
```diff
@@ -36,6 +36,8 @@ pub enum PostgresIo {
     ReadStream,
 }
 
+static VCHORDG_ENABLE: GucSetting<bool> = GucSetting::<bool>::new(true);
```
usamoi:
```suggestion
static VCHORDG_ENABLE_SCAN: GucSetting<bool> = GucSetting::<bool>::new(true);
```

src/index/gucs.rs
```diff
@@ -215,6 +235,10 @@ pub fn init() {
     }
 }
 
+pub fn vchordg_enable() -> bool {
```
usamoi:
```suggestion
pub fn vchordg_enable_scan() -> bool {
```

src/sql/finalize.sql
```diff
@@ -148,6 +148,76 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordrq_amhan
...
+    BEGIN
+        IF exact_search THEN
+            SET LOCAL vchordrq.enable_scan = off;
+        ELSE
+            IF rough_probes = '' THEN
+                accu_probes := '';
+            ELSIF position(',' in rough_probes) > 0 THEN
+                accu_probes := '9999,9999';
+            ELSE
+                accu_probes := '9999';
```
```suggestion
                accu_probes := '65535';
```

VoVAllen:
@cutecutecat @usamoi  mentioned there's limitation on sharedm memory size? Will it be a waste to store too much information?

src/sql/finalize.sql
```diff
@@ -148,6 +148,76 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordrq_amhan
...
+        END IF;
+        EXECUTE
+            format('SELECT array_agg(id) FROM (%s) AS result(id)', query)
+        INTO
+            accu;
+    EXCEPTION WHEN OTHERS THEN
+         RAISE EXCEPTION 'Error executing Ground Truth query "%": %', query, SQLERRM;
+    END;
+
+    accu_k := COALESCE(cardinality(accu), 0);
```
usamoi: `array_length` may return null, while `cardinality` does not.

VoVAllen:
And we will need to store the vector for each (table, column). How shall we design the storage? And the string name is variable length.

usamoi (approved):

PR #295, opened 2025-07-24
feat: record queries in internal table record - BGworker

cutecutecat:


src/bgworker/mq.rs
```diff
@@ -0,0 +1,214 @@
...
+    handle: NonNull<pg_sys::shm_mq_handle>,
+    creator: bool,
+}
+
+impl Drop for MessageQueueSender {
+    fn drop(&mut self) {
+        pgrx::warning!("Dropping MessageQueueSender, detaching handle");
+        unsafe {
+            if !self.creator {
+                pg_sys::pfree(self.handle.as_ptr() as _);
```
VoVAllen: I think for handle, it always need to be detach instead of pfree?

cutecutecat: > I think for handle, it always need to be detach instead of pfree?

If we detach a `handle`, the mq behind it is also detached, or mark as destroyed.

https://github.com/postgres/postgres/blob/e1c3654839e464957675344a1e949489d98b103b/src/backend/storage/ipc/shm_mq.c#L894

VoVAllen: Based on my understanding, Postgres manages memory here according to MemoryContext. This memory is automatically reclaimed when the MemoryContext is switched or destroyed, so manual calls to pfree might not be necessary. But I think it should be fine to call pfree here.

VoVAllen:
TODO: dump the recorded value to disk periodically like pg_wait_sampling, and load it when started if it exists?

crates/vector/src/vect.rs
```diff
@@ -231,3 +232,23 @@ impl<S: Floating> PartialOrd for VectBorrowed<'_, S> {
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "[")?;
+
+        let mut first = true;
+        for elem in self.slice().iter() {
+            if !first {
+                write!(f, ", ")?;
+            }
+            write!(f, "{elem:.2}")?;
```
usamoi: This will result in a loss of precision.

VoVAllen:
And also `recorder` -> `collector` sounds better to me

src/index/functions.rs
```diff
@@ -84,3 +88,56 @@ impl Drop for Index {
         }
     }
 }
+
+#[pgrx::pg_extern(sql = "")]
+fn _vchordrq_logged_queries(
+    indexrelid: Oid,
+) -> pgrx::iter::SetOfIterator<'static, pgrx::composite_type!('static, LOGGED_QUERY_COMPOSITE_TYPE)>
```
usamoi: Don't use a constant. I'm concerned that it doesn't work well with schema generation in pgrx.

src/index/functions.rs
```diff
@@ -84,3 +88,56 @@ impl Drop for Index {
...
+{
+    use pgrx::heap_tuple::PgHeapTuple;
+
+    let index_oid = if indexrelid == pgrx::pg_sys::InvalidOid {
+        None
+    } else {
+        Some(indexrelid.to_u32())
+    };
```
usamoi: Please forbid it. It may leads to leaking userdata.

src/sql/finalize.sql
```diff
@@ -188,12 +191,14 @@ BEGIN
         IF exact_search THEN
             SET LOCAL vchordrq.enable_scan = off;
         ELSE
-            IF rough_probes = '' THEN
-                accu_probes := '';
-            ELSIF position(',' in rough_probes) > 0 THEN
-                accu_probes := '65535,65535';
-            ELSE
-                accu_probes := '65535';
+            IF accu_probes = '' THEN
```
usamoi: A unique sentinel value is perferred. It avoids confusing user-provided arguments with default parameters.

src/sql/bootstrap.sql
```diff
@@ -4,3 +4,12 @@ CREATE TYPE scalar8;
...
 CREATE TYPE sphere_scalar8;
+
+CREATE TYPE logged_query AS (
+    table_schema TEXT,
+    table_name TEXT,
+    column_name TEXT,
+    operator TEXT,
+    vector_text TEXT,
+    simplified_query TEXT
+);
```
usamoi: A PostgreSQL cluster could contain multiple databases simultaneously, and indexes in different databases can have the same `oid`. It is essential to record the database oid. And it's necessary to record `table_schema`, `table_name`, `column_name`, `operator`. Please note that indexes can be built on expressions rather than just columns. Please note that you must call the `index_open` procedure to ensure the user has permission to access the specific index.

src/index/gucs.rs
```diff
@@ -167,6 +171,26 @@ pub fn init() {
...
+    GucRegistry::define_int_guc(
+        c"vchordrq.log_query_sample_rate",
+        c"`log_query_sample_rate` argument of vchordrq.",
+        c"`log_query_sample_rate` argument of vchordrq.",
+        &VCHORDRQ_LOG_QUERY_SAMPLE_RATE,
+        1,
+        100000,
+        GucContext::Userset,
+        GucFlags::default(),
+    );
```
usamoi: 1. Why are there 4 GUC variables, but only 2 are registered?
2. Please use a string GUC to ensure extensibility. Like this:
```
SET vchordrq.simpling TO '0.001, t_val_idx; 0.002, t_2_val_idx';
```

src/index/hook.rs
```diff
@@ -167,6 +184,13 @@ pub fn init() {
...
             Ordering::Relaxed,
         );
+        PREV_SHMEM_STARTUP_HOOK.store(
+            transmute::<Option<unsafe extern "C-unwind" fn() -> _>, *mut ()>(
+                pgrx::pg_sys::shmem_startup_hook,
+            ),
+            Ordering::Relaxed,
+        );
         pgrx::pg_sys::ExecutorStart_hook = Some(executor_start);
+        pgrx::pg_sys::shmem_startup_hook = Some(shmem_startup);
```
usamoi: 1. Please do not put unrelated code in the same file.
2. Why is there no `shmem_request_hook`?

src/index/vchordrq/scanners/default.rs
```diff
@@ -102,6 +120,18 @@ impl SearchBuilder for DefaultBuilder {
...
+            let mut rng = rand::rng();
+            if rng.random_bool(rate) {
+                let query = Query::new(self.table_oid, self.index_oid, opfamily, vector.clone());
+                if let Some(q) = query {
+                    unsafe {
+                        QueryRecorderMaster::push(q);
+                    }
+                }
+            }
+        }
```
usamoi: 1. Don't read the variable like this here. Put it inside `SearchOptions`.
2. Don't put the logic here. Use a callback to separate the PostgreSQL shim from the logic.

src/index/vchordrq/scanners/default.rs
```diff
@@ -34,6 +37,18 @@ pub struct DefaultBuilder {
...
+}
+
+impl DefaultBuilder {
+    pub fn set_index_oid(&mut self, oid: u32) {
+        self.index_oid = oid;
+    }
+
+    pub fn set_table_oid(&mut self, oid: u32) {
+        self.table_oid = oid;
+    }
```
usamoi: No.

src/index/vchordrq/scanners/default.rs
```diff
@@ -53,6 +68,8 @@ impl SearchBuilder for DefaultBuilder {
             opfamily,
             orderbys: Vec::new(),
             spheres: Vec::new(),
+            index_oid: u32::default(),
+            table_oid: u32::default(),
```
usamoi: Delete it.

src/index/vchordrq/am/mod.rs
```diff
@@ -460,6 +460,8 @@ pub unsafe extern "C-unwind" fn amrescan(
             | Opfamily::HalfvecIp
             | Opfamily::HalfvecCosine => {
                 let mut builder = DefaultBuilder::new(opfamily);
+                builder.set_table_oid((*(*scan).heapRelation).rd_id.to_u32());
+                builder.set_index_oid((*(*scan).indexRelation).rd_id.to_u32());
```
usamoi: Delete.

src/recorder/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+pub use types::Query;
+use types::{COMMAND_REQUEST, Command, MAX_QUERY_LEN, QUERY_LOAD_MQ, QUERY_PUSH_MQ, VCHORD_MAGIC};
+pub use worker::QueryRecorderMaster;
+use worker::{QueryRecorderWorker, set_command};
+
+mod mqueue;
+mod types;
+mod worker;
+
+pub unsafe fn init() {
```
usamoi: It doesn't seem necessary to mark it as unsafe.

src/recorder/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    use pgrx::bgworkers::{BackgroundWorkerBuilder, BgWorkerStartTime};
+    use std::time::Duration;
+    BackgroundWorkerBuilder::new("vectors")
```
usamoi: It would be better if we use a different name.

src/recorder/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+            QUERY_LOAD_MQ = pgrx::pg_sys::shm_toc_lookup(toc, 1, false) as *mut _;
+            COMMAND_REQUEST = pgrx::pg_sys::shm_toc_lookup(toc, 2, false) as *mut _;
+        }
+    }
+}
+
+pub unsafe extern "C-unwind" fn handle_sigterm(_postgres_signal_arg: i32) {
+    pgrx::warning!("Recorder worker: Received SIGTERM, shutting down background worker");
+    unsafe { set_command(Command::Shutdown) };
+}
```
usamoi: It's not guarded.

src/index/functions.rs
```diff
@@ -84,3 +88,56 @@ impl Drop for Index {
...
+                    index_cache.insert(query.index_oid, column.clone());
+                    column
+                }
+                None => continue,
+            },
+        };
+        dumps.push(query.dump(namespace, table_name, column_name));
+    }
+    pgrx::iter::SetOfIterator::new(dumps.into_iter().map(|dump| {
+        let mut view = PgHeapTuple::new_composite_type(LOGGED_QUERY_COMPOSITE_TYPE).unwrap();
```
usamoi: This will search for the type in the current `search_path`, so it is not quite correct.

usamoi (changes requested):
would check 3 files left later

src/collector/hook.rs
```diff
@@ -0,0 +1,152 @@
...
+use crate::collector::QueryCollectorMaster;
+use crate::collector::types::{
+    COMMAND_REQUEST, Command, MAX_QUERY_LEN, QUERY_LOAD_MQ, QUERY_PUSH_MQ, VCHORD_MAGIC,
+};
+use std::sync::atomic::AtomicPtr;
+
+pub const COMMAND_STRUCT_SIZE: usize =
+    size_of::<Command>().next_multiple_of(pgrx::pg_sys::ALIGNOF_BUFFER as _);
+
+#[cfg(not(any(feature = "pg13", feature = "pg14")))]
```
usamoi:
```suggestion
#[cfg(any(feature = "pg15", feature = "pg16", feature = "pg17", feature = "pg18"))]
```

src/collector/hook.rs
```diff
@@ -0,0 +1,152 @@
...
+            QueryCollectorMaster::drop(pgrx::pg_sys::MyDatabaseId.to_u32(), object_id.to_u32());
+        }
+    }
+}
+
+pub fn init() {
+    unsafe {
+        use core::mem::transmute;
+        use std::sync::atomic::Ordering;
+        #[cfg(not(any(feature = "pg13", feature = "pg14")))]
```
usamoi:
```suggestion
        #[cfg(any(feature = "pg15", feature = "pg16", feature = "pg17", feature = "pg18"))]
```

src/collector/hook.rs
```diff
@@ -0,0 +1,152 @@
...
+
+unsafe fn shm_estimate_size() -> pgrx::pg_sys::Size {
+    let mut estimator = pgrx::pg_sys::shm_toc_estimator {
+        space_for_chunks: COMMAND_STRUCT_SIZE + MAX_QUERY_LEN * 2,
+        number_of_keys: 3,
+    };
+    unsafe { pgrx::pg_sys::shm_toc_estimate(&mut estimator) }
+}
+
+#[cfg(not(any(feature = "pg13", feature = "pg14")))]
```
usamoi:
```suggestion
#[cfg(any(feature = "pg15", feature = "pg16", feature = "pg17", feature = "pg18"))]
```

src/collector/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+#[pgrx::pg_guard]
+#[unsafe(no_mangle)]
+pub extern "C-unwind" fn _query_logger_main(_arg: pgrx::pg_sys::Datum) {
+    use core::mem::transmute;
+    use pgrx::pg_sys;
+    unsafe {
+        pg_sys::pqsignal(pg_sys::SIGTERM as i32, Some(handle_sigterm));
+        pg_sys::pqsignal(
+            pg_sys::SIGHUP as i32,
+            transmute::<*mut (), pg_sys::pqsigfunc>(pg_sys::SignalHandlerForConfigReload as _),
```
usamoi: Doesn't it work?
```suggestion
            Some(pg_sys::SignalHandlerForConfigReload),
```

src/collector/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+        );
+        pg_sys::BackgroundWorkerUnblockSignals();
+        let mut worker = QueryCollectorWorker::new();
+        worker.run();
+    }
+}
+
+#[pgrx::pg_guard]
+pub unsafe extern "C-unwind" fn handle_sigterm(_postgres_signal_arg: i32) {
+    pgrx::warning!("Collector worker: Received SIGTERM, shutting down background worker");
```
usamoi: I guess it's not async-signal-safe?

src/collector/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+    use pgrx::pg_sys;
+    unsafe {
+        pg_sys::pqsignal(pg_sys::SIGTERM as i32, Some(handle_sigterm));
+        pg_sys::pqsignal(
+            pg_sys::SIGHUP as i32,
+            transmute::<*mut (), pg_sys::pqsigfunc>(pg_sys::SignalHandlerForConfigReload as _),
+        );
+        pg_sys::pqsignal(
+            pg_sys::SIGUSR1 as i32,
+            transmute::<*mut (), pg_sys::pqsigfunc>(pg_sys::procsignal_sigusr1_handler as _),
```
usamoi, line 49: Doesn't it work?

```suggestion
            Some(pg_sys::procsignal_sigusr1_handler),
```

src/collector/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+    hook::init()
+}
+
+#[pgrx::pg_guard]
+#[unsafe(no_mangle)]
+pub extern "C-unwind" fn _query_logger_main(_arg: pgrx::pg_sys::Datum) {
+    use core::mem::transmute;
+    use pgrx::pg_sys;
+    unsafe {
+        pg_sys::pqsignal(pg_sys::SIGTERM as i32, Some(handle_sigterm));
```
usamoi, line 45: We prefer `as _` because on different platforms, its type is different.

src/index/functions.rs
```diff
@@ -84,3 +87,74 @@ impl Drop for Index {
...
+    indexrelid: Oid,
+) -> TableIterator<
+    'static,
+    (
+        pgrx::name!(schema_name, String),
+        pgrx::name!(table_name, String),
+        pgrx::name!(column_name, Option<String>),
+        pgrx::name!(operator, String),
+        pgrx::name!(vector_text, String),
+        pgrx::name!(simplified_query, Option<String>),
```
usamoi, line 102: We can return only `vector_text` and compute the other content using a view.

src/index/gucs.rs
```diff
@@ -36,6 +36,10 @@ pub enum PostgresIo {
     ReadStream,
 }
 
+static VCHORDRQ_MAX_LOGGED_QUERIES_PER_INDEX: GucSetting<i32> = GucSetting::<i32>::new(0);
```
usamoi, line 39: All GUCs apply to a single index only. No need to add `PER_INDEX`.

src/index/gucs.rs
```diff
@@ -36,6 +36,10 @@ pub enum PostgresIo {
...
+
+static VCHORDRQ_LOG_QUERIES_SAMPLE_RATE: GucSetting<f64> = GucSetting::<f64>::new(0.01);
```
usamoi, line 41: It’s better to set it to `0.0` and let the user decide, rather than providing a default value without any justification.

src/index/gucs.rs
```diff
@@ -167,6 +171,26 @@ pub fn init() {
         GucContext::Userset,
         GucFlags::default(),
     );
+    GucRegistry::define_int_guc(
+        c"vchordrq.max_logged_queries_per_index",
+        c"`max_logged_queries_per_index` argument of vchordrq.",
+        c"`max_logged_queries_per_index` argument of vchordrq.",
+        &VCHORDRQ_MAX_LOGGED_QUERIES_PER_INDEX,
+        0,
+        10000,
```
usamoi, line 180: We can set it to a larger value, like 65535, right?

src/index/vchordrq/am/mod.rs
```diff
@@ -434,6 +435,8 @@ pub unsafe extern "C-unwind" fn amrescan(
             io_search: gucs::vchordrq_io_search(),
             io_rerank: gucs::vchordrq_io_rerank(),
             prefilter: gucs::vchordrq_prefilter(),
+            collect_enable: gucs::vchordrq_max_logged_queries_per_index() > 0,
+            collect_rate: gucs::vchordrq_log_queries_sample_rate(),
```
usamoi, line 443: 1. Make it an `Option`.
2. Let's enable it by `sample_rate > 0.0` is better.

src/index/vchordrq/am/mod.rs
```diff
@@ -450,6 +453,12 @@ pub unsafe extern "C-unwind" fn amrescan(
                 )
             })
         };
+        let sender = DefaultSender {
+            database_oid: (*(*scan).heapRelation).rd_lockInfo.lockRelId.dbId.to_u32(),
```
usamoi, line 461: Doesn't `MyDatabaseOid` work?

src/index/collector.rs
```diff
@@ -0,0 +1,52 @@
...
+// Copyright (c) 2025 TensorChord Inc.
+
+use crate::collector::{Query, QueryCollectorMaster};
+use crate::index::vchordrq::opclass::Opfamily;
+use crate::index::vchordrq::scanners::SearchOptions;
+use rand::Rng;
+use vchordrq::types::OwnedVector;
+
+pub trait CollectorSender {
+    fn exec(&self, options: &SearchOptions, vector: &OwnedVector);
```
usamoi, line 22:
```suggestion
    fn send(&self, vector: &OwnedVector);
```

1. If a value is used by the *Sender*, it shouldn't be placed in `SearchOptions`. If a value is used by the `Scanner`, it shouldn't be accessible to the `Sender`.
2. What about using a string? It's compatible with maxsim.

src/collector/worker.rs
```diff
@@ -0,0 +1,266 @@
...
+    BgWorkerLockGuard, COMMAND_REQUEST, Command, QUERY_LOAD_MQ, QUERY_PUSH_MQ, Query, VCHORD_MAGIC,
+    WorkerState,
+};
+use crate::index::gucs;
+use std::collections::{HashMap, VecDeque};
+
+pub const MULTI_ACCESS_LOCK: u32 = 0;
+pub const BACKGROUND_WORKER_LOCK: u32 = 1;
+
+pub struct QueryCollectorMaster {}
```
usamoi, line 23: This is a meaningless type, created solely to make the call look more "OOP".

PR #312, opened 2025-08-12
docker: download vchord with wget in build

usamoi:


VoVAllen:
Why? I think it's intentional to have a binary docker image.

usamoi:
> Why? I think it's intentional to have a binary docker image.

`COPY --from=binary` causes the image size to permanently increase for `vchord.deb`. If someone really wants to do this, they can also download the debian package using `ADD` with a URL.

PR #321, opened 2025-08-25
feat: record queries by sqlite

cutecutecat:
job: +psql_windows
job: +psql_macos
job: +check_debian
job: +psql_alpine

## GUC
* `vchordrq.query_sampling_enable`: Whether to enable the query logger
  * bool, default off
* `vchordrq.query_sampling_max_records`: Maximum logged queries per index
  * int, 0 to **10000**, default 0
* `vchordrq.query_sampling_rate`: Sample rate of query to be logged
  * float, 0 to 1, default 0.01 (for 1% pick rate)

## Function
* `vchordrq_sampled_queries(index: Oid)`: Show logged queries for the given index
* `vchordrq_sampled_vectors(index: Oid)`: Get logged vector samples (only for internal use).

## View
* `vchordrq_sampled_queries`: Show logged queries for all indexes which you have access

## Example
```sql
-- Preset
CREATE TABLE t (id SERIAL PRIMARY KEY, val vector(3));
INSERT INTO t (val)
SELECT ARRAY[i * 0.0001, i * 0.00005, i * 0.0002]::vector(3) FROM generate_series(1, 10000) as s(i);
CREATE INDEX idx ON t USING vchordrq (val vector_l2_ops);
CREATE TABLE t1 (val vector(4), id SERIAL PRIMARY KEY);
INSERT INTO t1 (val)
SELECT ARRAY[i * 0.0001, i * 0.00005, i * 0.0002, i * 0.001]::vector(4) FROM generate_series(1, 10000) as s(i);
CREATE INDEX idx1 ON t1 USING vchordrq (val vector_l2_ops);

SET vchordrq.query_sampling_enable = on;
-- Log up to 2 queries
SET vchordrq.query_sampling_max_records = 2;
-- Record query at 100% rate
SET vchordrq.query_sampling_rate = 1;

-- Record
SELECT * from t ORDER BY val <-> '[0.50, 0.25, 1.00]';
SELECT val from t1 ORDER BY val <-> '[0.50, 0.25, 1.00, 0]';

SELECT * from vchordrq_sampled_queries('idx1');

-- Result
 schema_name | table_name | column_name | operator |            vector_text             
-------------+------------+-------------+----------+------------------------------------
 public      | t1         | val         | <->      | [0.5,0.25,1,0]

SELECT * from vchordrq_sampled_queries;

-- Result
 schema_name | table_name | column_name | operator |            vector_text             
-------------+------------+-------------+----------+------------------------------------
 public      | t          | val         | <->      | [0.5,0.25,1]
 public      | t1         | val         | <->      | [0.5,0.25,1,0]

-- Recall metric
SELECT AVG(recall_value)
FROM (
    SELECT
        vchordrq_evaluate_query_recall(
            query => format(
                'SELECT ctid FROM %I.%I ORDER BY %I %s ''%s'' LIMIT 10',
                lq.schema_name,
                lq.table_name,
                lq.column_name,
                lq.operator,
                lq.vector_text
            )
        ) AS recall_value
    FROM
        vchordrq_sampled_queries('idx') AS lq
) AS eval_results;

-- Result
 avg 
-----
   1
```

## Restriction
- Only users with **read access** to the index can obtain the data of the corresponding index
- Now, the record for `vchordg index` is not enabled

## Internal
Structure
- Use a bunded sqlite database provided by https://github.com/rusqlite/rusqlite
- `INIT`: Connect and create a disk-based `sqlite table` at `PG_init`
- `INSERT`: Insert to `sqlite table` when query logged, and maintain max rows to `max_logged_queries_per_index` for **each index**
- `DELETE`: Delete rows belongs to an index from `sqlite table` when a `DROP TABLE` or `DROP INDEX` triggered hook
- `READ`: Read all rows belongs to an PG index when the function `vchordrq_logged_queries` called

Synchronous
- There will only be 1 read/delete/insert request sent to sqlite at a time
- `INSERT`: If cannot get the lock, then skip
- `READ/DELETE`:  If cannot get the lock, then wait

## Latency
- For first logged query (with initialize): +30ms
- For following logged queries (without initialize): +10ms

### DIFF
--- a/src/index/functions.rs
+++ b/src/index/functions.rs
@@ -13,6 +13,8 @@
 // Copyright (c) 2025 TensorChord Inc.
 
 use crate::index::storage::PostgresRelation;
+use crate::recorder::dump;
+use pgrx::iter::SetOfIterator;
 use pgrx::pg_sys::Oid;
 use pgrx_catalog::{PgAm, PgClass, PgClassRelkind};
 
@@ -83,4 +85,26 @@ impl Drop for Index {
             pgrx::pg_sys::index_close(self.raw, self.lockmode);
         }
     }
+}
+
+#[pgrx::pg_extern(sql = "")]
+fn _vchordrq_sampled_vectors(indexrelid: Oid) -> SetOfIterator<'static, String> {
+    let pg_am = PgAm::search_amname(c"vchordrq").unwrap();
+    let Some(pg_am) = pg_am.get() else {
+        pgrx::error!("vchord is not installed");
+    };
+    let pg_class = PgClass::search_reloid(indexrelid).unwrap();
+    let Some(pg_class) = pg_class.get() else {
+        pgrx::error!("the relation does not exist");
+    };
+    if pg_class.relkind() != PgClassRelkind::Index {
+        pgrx::error!("the relation {:?} is not an index", pg_class.relname());
+    }
+    if pg_class.relam() != pg_am.oid() {
+        pgrx::error!("the index {:?} is not a vchordrq index", pg_class.relname());
+    }
+    // The user must have access to the index, if not, raise an error from Postgres.
+    let _relation = Index::open(indexrelid, pgrx::pg_sys::AccessShareLock as _);
+    let queries = dump(indexrelid.to_u32());
+    SetOfIterator::new(queries)
 }
--- a/src/index/gucs.rs
+++ b/src/index/gucs.rs
@@ -27,6 +27,12 @@ pub enum PostgresIo {
     ReadStream,
 }
 
+static VCHORDRQ_QUERY_SAMPLING_ENABLE: GucSetting<bool> = GucSetting::<bool>::new(false);
+
+static VCHORDRQ_QUERY_SAMPLING_MAX_RECORDS: GucSetting<i32> = GucSetting::<i32>::new(0);
+
+static VCHORDRQ_QUERY_SAMPLING_RATE: GucSetting<f64> = GucSetting::<f64>::new(0.0);
+
 static VCHORDG_ENABLE_SCAN: GucSetting<bool> = GucSetting::<bool>::new(true);
 
 static VCHORDG_EF_SEARCH: GucSetting<i32> = GucSetting::<i32>::new(64);
@@ -155,6 +161,34 @@ pub fn init() {
         c"`io_rerank` argument of vchordrq.",
         c"`io_rerank` argument of vchordrq.",
         &VCHORDRQ_IO_RERANK,
+        GucContext::Userset,
+        GucFlags::default(),
+    );
+    GucRegistry::define_bool_guc(
+        c"vchordrq.query_sampling_enable",
+        c"`query_sampling_enable` argument of vchordrq.",
+        c"`query_sampling_enable` argument of vchordrq.",
+        &VCHORDRQ_QUERY_SAMPLING_ENABLE,
+        GucContext::Userset,
+        GucFlags::default(),
+    );
+    GucRegistry::define_int_guc(
+        c"vchordrq.query_sampling_max_records",
+        c"`query_sampling_max_records` argument of vchordrq.",
+        c"`query_sampling_max_records` argument of vchordrq.",
+        &VCHORDRQ_QUERY_SAMPLING_MAX_RECORDS,
+        0,
+        10000,
+        GucContext::Userset,
+        GucFlags::default(),
+    );
+    GucRegistry::define_float_guc(
+        c"vchordrq.query_sampling_rate",
+        c"`query_sampling_rate` argument of vchordrq.",
+        c"`query_sampling_rate` argument of vchordrq.",
+        &VCHORDRQ_QUERY_SAMPLING_RATE,
+        0.0,
+        1.0,
         GucContext::Userset,
         GucFlags::default(),
     );
@@ -330,4 +364,16 @@ pub fn vchordrq_io_rerank() -> Io {
         #[cfg(any(feature = "pg17", feature = "pg18"))]
         PostgresIo::ReadStream => Io::Stream,
     }
+}
+
+pub fn vchordrq_query_sampling_enable() -> bool {
+    VCHORDRQ_QUERY_SAMPLING_ENABLE.get()
+}
+
+pub fn vchordrq_query_sampling_max_records() -> u32 {
+    VCHORDRQ_QUERY_SAMPLING_MAX_RECORDS.get() as u32
+}
+
+pub fn vchordrq_query_sampling_rate() -> f64 {
+    VCHORDRQ_QUERY_SAMPLING_RATE.get()
 }
--- a/src/index/vchordg/scanners/default.rs
+++ b/src/index/vchordg/scanners/default.rs
@@ -18,6 +18,7 @@ use crate::index::scanners::{Io, SearchBuilder};
 use crate::index::vchordg::algo::*;
 use crate::index::vchordg::opclass::Opfamily;
 use crate::index::vchordg::scanners::SearchOptions;
+use crate::recorder::{Recorder, halfvec_out, vector_out};
 use algo::accessor::{Dot, L2S};
 use algo::*;
 use distance::Distance;
@@ -26,6 +27,7 @@ use std::num::NonZero;
 use vchordg::operator::{self};
 use vchordg::types::{DistanceKind, OwnedVector, VectorKind};
 use vchordg::*;
+use vector::VectorOwned;
 use vector::vect::{VectBorrowed, VectOwned};
 
 pub struct DefaultBuilder {
@@ -78,6 +80,7 @@ impl SearchBuilder for DefaultBuilder {
         options: SearchOptions,
         _fetcher: impl Fetcher + 'b,
         bump: &'b impl Bump,
+        recorder: impl Recorder,
     ) -> Box<dyn Iterator<Item = (f32, [u16; 3], bool)> + 'b>
     where
         R: RelationRead + RelationPrefetch + RelationReadStream,
@@ -120,7 +123,7 @@ impl SearchBuilder for DefaultBuilder {
             match (opfamily.vector_kind(), opfamily.distance_kind()) {
                 (VectorKind::Vecf32, DistanceKind::L2S) => {
                     type Op = operator::Op<VectOwned<f32>, L2S>;
-                    let unprojected = if let OwnedVector::Vecf32(vector) = vector {
+                    let unprojected = if let OwnedVector::Vecf32(vector) = vector.clone() {
                         VectBorrowed::new(bump.alloc_slice(vector.slice()))
                     } else {
                         unreachable!()
@@ -215,7 +218,7 @@ impl SearchBuilder for DefaultBuilder {
                 }
                 (VectorKind::Vecf16, DistanceKind::L2S) => {
                     type Op = operator::Op<VectOwned<f16>, L2S>;
-                    let unprojected = if let OwnedVector::Vecf16(vector) = vector {
+                    let unprojected = if let OwnedVector::Vecf16(vector) = vector.clone() {
                         VectBorrowed::new(bump.alloc_slice(vector.slice()))
                     } else {
                         unreachable!()
@@ -310,7 +313,7 @@ impl SearchBuilder for DefaultBuilder {
                 }
                 (VectorKind::Vecf32, DistanceKind::Dot) => {
                     type Op = operator::Op<VectOwned<f32>, Dot>;
-                    let unprojected = if let OwnedVector::Vecf32(vector) = vector {
+                    let unprojected = if let OwnedVector::Vecf32(vector) = vector.clone() {
                         VectBorrowed::new(bump.alloc_slice(vector.slice()))
                     } else {
                         unreachable!()
@@ -405,7 +408,7 @@ impl SearchBuilder for DefaultBuilder {
                 }
                 (VectorKind::Vecf16, DistanceKind::Dot) => {
                     type Op = operator::Op<VectOwned<f16>, Dot>;
-                    let unprojected = if let OwnedVector::Vecf16(vector) = vector {
+                    let unprojected = if let OwnedVector::Vecf16(vector) = vector.clone() {
                         VectBorrowed::new(bump.alloc_slice(vector.slice()))
                     } else {
                         unreachable!()
@@ -509,6 +512,16 @@ impl SearchBuilder for DefaultBuilder {
         } else {
             iter
         };
+        if recorder.is_enabled() {
+            match &vector {
+                OwnedVector::Vecf32(v) => {
+                    recorder.send(&vector_out(v.as_borrowed()));
+                }
+                OwnedVector::Vecf16(v) => {
+                    recorder.send(&halfvec_out(v.as_borrowed()));
+                }
+            }
+        }
         Box::new(iter.map(move |(distance, pointer)| {
             let (key, _) = pointer_to_kv(pointer);
             (opfamily.output(distance), key, recheck)
--- a/src/index/vchordrq/scanners/default.rs
+++ b/src/index/vchordrq/scanners/default.rs
@@ -19,6 +19,7 @@ use crate::index::vchordrq::algo::*;
 use crate::index::vchordrq::filter::filter;
 use crate::index::vchordrq::opclass::Opfamily;
 use crate::index::vchordrq::scanners::SearchOptions;
+use crate::recorder::{Recorder, halfvec_out, vector_out};
 use algo::accessor::{Dot, L2S};
 use algo::prefetcher::*;
 use algo::*;
@@ -82,6 +83,7 @@ impl SearchBuilder for DefaultBuilder {
         options: SearchOptions,
         mut fetcher: impl Fetcher + 'b,
         bump: &'b impl Bump,
+        recorder: impl Recorder,
     ) -> Box<dyn Iterator<Item = (f32, [u16; 3], bool)> + 'b>
     where
         R: RelationRead + RelationPrefetch + RelationReadStream,
@@ -120,7 +122,7 @@ impl SearchBuilder for DefaultBuilder {
             match (opfamily.vector_kind(), opfamily.distance_kind()) {
                 (VectorKind::Vecf32, DistanceKind::L2S) => {
                     type Op = operator::Op<VectOwned<f32>, L2S>;
-                    let unprojected = if let OwnedVector::Vecf32(vector) = vector {
+                    let unprojected = if let OwnedVector::Vecf32(vector) = vector.clone() {
                         vector
                     } else {
                         unreachable!()
@@ -260,7 +262,7 @@ impl SearchBuilder for DefaultBuilder {
                 }
                 (VectorKind::Vecf32, DistanceKind::Dot) => {
                     type Op = operator::Op<VectOwned<f32>, Dot>;
-                    let unprojected = if let OwnedVector::Vecf32(vector) = vector {
+                    let unprojected = if let OwnedVector::Vecf32(vector) = vector.clone() {
                         vector
                     } else {
                         unreachable!()
@@ -400,7 +402,7 @@ impl SearchBuilder for DefaultBuilder {
                 }
                 (VectorKind::Vecf16, DistanceKind::L2S) => {
                     type Op = operator::Op<VectOwned<f16>, L2S>;
-                    let unprojected = if let OwnedVector::Vecf16(vector) = vector {
+                    let unprojected = if let OwnedVector::Vecf16(vector) = vector.clone() {
                         vector
                     } else {
                         unreachable!()
@@ -540,7 +542,7 @@ impl SearchBuilder for DefaultBuilder {
                 }
                 (VectorKind::Vecf16, DistanceKind::Dot) => {
                     type Op = operator::Op<VectOwned<f16>, Dot>;
-                    let unprojected = if let OwnedVector::Vecf16(vector) = vector {
+                    let unprojected = if let OwnedVector::Vecf16(vector) = vector.clone() {
                         vector
                     } else {
                         unreachable!()
@@ -689,6 +691,16 @@ impl SearchBuilder for DefaultBuilder {
         } else {
             iter
         };
+        if recorder.is_enabled() {
+            match &vector {
+                OwnedVector::Vecf32(v) => {
+                    recorder.send(&vector_out(v.as_borrowed()));
+                }
+                OwnedVector::Vecf16(v) => {
+                    recorder.send(&halfvec_out(v.as_borrowed()));
+                }
+            }
+        }
         Box::new(iter.map(move |(distance, pointer)| {
             let (key, _) = pointer_to_kv(pointer);
             (distance, key, recheck)
--- a/src/index/vchordrq/scanners/maxsim.rs
+++ b/src/index/vchordrq/scanners/maxsim.rs
@@ -18,6 +18,7 @@ use crate::index::vchordrq::algo::*;
 use crate::index::vchordrq::filter::filter;
 use crate::index::vchordrq::opclass::Opfamily;
 use crate::index::vchordrq::scanners::SearchOptions;
+use crate::recorder::Recorder;
 use algo::accessor::Dot;
 use algo::prefetcher::*;
 use algo::*;
@@ -72,6 +73,7 @@ impl SearchBuilder for MaxsimBuilder {
         options: SearchOptions,
         mut fetcher: impl Fetcher + 'b,
         bump: &'b impl Bump,
+        _sender: impl Recorder,
     ) -> Box<dyn Iterator<Item = (f32, [u16; 3], bool)> + 'b>
     where
         R: RelationRead + RelationPrefetch + RelationReadStream,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,6 +18,7 @@
 
 mod datatype;
 mod index;
+mod recorder;
 mod upgrade;
 
 pgrx::pg_module_magic!(
@@ -53,6 +54,7 @@ unsafe extern "C-unwind" fn _pg_init() {
     }
     IS_MAIN.set(true);
     index::init();
+    recorder::init();
     unsafe {
         #[cfg(any(feature = "pg13", feature = "pg14"))]
         pgrx::pg_sys::EmitWarningsOnPlaceholders(c"vchord".as_ptr());
--- a/src/sql/finalize.sql
+++ b/src/sql/finalize.sql
@@ -142,6 +142,102 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchord_vector_
 CREATE FUNCTION quantize_to_scalar8(halfvec) RETURNS scalar8
 IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchord_halfvec_quantize_to_scalar8_wrapper';
 
+CREATE FUNCTION vchordrq_sampled_vectors(regclass)
+RETURNS SETOF TEXT
+STRICT LANGUAGE c AS 'MODULE_PATHNAME', '_vchordrq_sampled_vectors_wrapper';
+
+CREATE OR REPLACE FUNCTION vchordrq_sampled_queries(regclass)
+RETURNS TABLE(
+    schema_name NAME,
+    table_name NAME,
+    column_name NAME,
+    operator TEXT,
+    vector_text TEXT
+)
+LANGUAGE plpgsql
+STRICT AS $$
+DECLARE
+    ext_schema TEXT;
+    query_text TEXT;
+BEGIN
+    SELECT n.nspname
+    INTO ext_schema
+    FROM pg_catalog.pg_extension e
+    JOIN pg_catalog.pg_namespace n ON n.oid = e.extnamespace
+    WHERE e.extname = 'vchord';
+
+    IF ext_schema IS NULL THEN
+        RAISE EXCEPTION 'vchord is not installed';
+    END IF;
+
+    query_text := format(
+        $q$
+        WITH index_metadata AS (
+            SELECT
+                NS.nspname AS schema_name,
+                C.relname AS table_name,
+                PA.attname AS column_name,
+                CASE
+                    WHEN OP.opcname LIKE '%%l2%%' THEN '<->'
+                    WHEN OP.opcname LIKE '%%ip%%' THEN '<#>'
+                    WHEN OP.opcname LIKE '%%cosine%%' THEN '<=>'
+                    ELSE ''
+                END AS operator
+            FROM
+                pg_catalog.pg_index X
+            JOIN
+                pg_catalog.pg_class C ON C.oid = X.indrelid
+            JOIN
+                pg_catalog.pg_namespace NS ON C.relnamespace = NS.oid
+            JOIN
+                pg_catalog.pg_class I ON I.oid = X.indexrelid
+            JOIN
+                pg_catalog.pg_am A ON A.oid = I.relam
+            LEFT JOIN
+                pg_catalog.pg_opclass AS OP ON OP.oid = X.indclass[0]
+            LEFT JOIN
+                pg_catalog.pg_attribute PA ON PA.attrelid = X.indrelid AND PA.attnum = X.indkey[0]
+            WHERE
+                A.amname = 'vchordrq'
+                AND C.relkind = 'r'
+                AND X.indnatts = 1
+                AND X.indexrelid = %1$s
+        )
+        SELECT
+            im.schema_name,
+            im.table_name,
+            im.column_name,
+            im.operator,
+            s.vector_text
+        FROM
+            index_metadata im,
+            LATERAL %2$I.vchordrq_sampled_vectors(%1$s) AS s(vector_text);
+        $q$,
+        $1::oid,
+        ext_schema
+    );
+    RETURN QUERY EXECUTE query_text;
+END;
+$$;
+
+CREATE VIEW vchordrq_sampled_queries AS
+SELECT
+    record.schema_name,
+    record.table_name,
+    record.column_name,
+    record.operator,
+    record.vector_text
+FROM
+    (
+        SELECT i.oid
+        FROM pg_catalog.pg_class AS i
+        JOIN pg_catalog.pg_index AS ix ON i.oid = ix.indexrelid
+        JOIN pg_catalog.pg_opclass AS opc ON ix.indclass[0] = opc.oid
+        JOIN pg_catalog.pg_am AS am ON opc.opcmethod = am.oid
+        WHERE am.amname = 'vchordrq'
+    ) AS index_oids
+CROSS JOIN LATERAL vchordrq_sampled_queries(index_oids.oid::regclass) AS record;
+
 CREATE FUNCTION vchordrq_amhandler(internal) RETURNS index_am_handler
 IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchordrq_amhandler_wrapper';
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -421,6 +421,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "fallible-iterator"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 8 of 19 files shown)


src/index/gucs.rs
```diff
@@ -27,6 +27,10 @@ pub enum PostgresIo {
     ReadStream,
 }
 
+static VCHORDRQ_MAX_LOGGED_QUERIES_PER_INDEX: GucSetting<i32> = GucSetting::<i32>::new(0);
+
+static VCHORDRQ_LOG_QUERIES_SAMPLE_RATE: GucSetting<f64> = GucSetting::<f64>::new(0.0);
```
usamoi: We’d better add a GUC to explicitly enable it.

src/index/functions.rs
```diff
@@ -84,3 +87,74 @@ impl Drop for Index {
...
+) -> TableIterator<
+    'static,
+    (
+        pgrx::name!(schema_name, String),
+        pgrx::name!(table_name, String),
+        pgrx::name!(column_name, Option<String>),
+        pgrx::name!(operator, String),
+        pgrx::name!(vector_text, String),
+        pgrx::name!(simplified_query, Option<String>),
+    ),
```
usamoi: Most of this is redundant. We should keep this function simple and retrieve all other information through the view. This function should returns `SETOF text`.

src/index/vchordrq/scanners/default.rs
```diff
@@ -108,6 +110,7 @@ impl SearchBuilder for DefaultBuilder {
         let Some(vector) = vector else {
             return Box::new(std::iter::empty()) as Box<dyn Iterator<Item = (f32, [u16; 3], bool)>>;
         };
+        sender.send_vchordrq(opfamily, vector.clone());
```
usamoi: We should defer it until after `default_search` is called, so that queries which should trigger errors are not recorded.

src/collector/worker.rs
```diff
@@ -0,0 +1,269 @@
...
+use std::path::Path;
+use std::sync::Mutex;
+
+const SQLITE_DB_PATH: &str = "vectorchord/collector.db";
+const SQLITE_DATABASE: Option<&CStr> = Some(c"main");
+const SQLITE_TABLE: &CStr = c"collector";
+
+pub static COLLECTOR: Mutex<Option<Connection>> = Mutex::new(None);
+
+pub struct QueryCollector {}
```
usamoi: Remove this type. It's confusing.

src/collector/worker.rs
```diff
@@ -0,0 +1,269 @@
...
+use crate::collector::Query;
+use rusqlite::Connection;
+use std::ffi::CStr;
+use std::fs;
+use std::path::Path;
+use std::sync::Mutex;
+
+const SQLITE_DB_PATH: &str = "vectorchord/collector.db";
+const SQLITE_DATABASE: Option<&CStr> = Some(c"main");
+const SQLITE_TABLE: &CStr = c"collector";
```
usamoi: 1. I suggest using a separate file for each PostgreSQL database, just like PG itself does.
2. And using a separate table for each PostgreSQL table, just like PG itself does.
3. I wonder if it works well with `pg_basebackup`: these files should not be copied by PostgreSQL backuper.

src/collector/worker.rs
```diff
@@ -0,0 +1,269 @@
...
+
+pub static COLLECTOR: Mutex<Option<Connection>> = Mutex::new(None);
```
usamoi: It's a joke, right? `Mutex` protects nothing.

src/collector/worker.rs
```diff
@@ -0,0 +1,269 @@
...
+impl QueryCollector {
+    pub fn init() {
+        let init_statement = "
+            CREATE TABLE IF NOT EXISTS collector (
+                database_oid INTEGER,
+                table_oid INTEGER,
+                index_oid INTEGER,
+                operator TEXT,
+                vector TEXT,
+                create_at REAL
```
usamoi:
```suggestion
                sample TEXT
```

src/collector/worker.rs
```diff
@@ -0,0 +1,269 @@
...
+//
+// Copyright (c) 2025 TensorChord Inc.
+use crate::collector::Query;
+use rusqlite::Connection;
+use std::ffi::CStr;
+use std::fs;
+use std::path::Path;
+use std::sync::Mutex;
+
+const SQLITE_DB_PATH: &str = "vectorchord/collector.db";
```
usamoi: Maybe we could use a different name `pg_sample` to avoid some people uninstalling the extension due to OCD triggered by inconsistent naming.

src/collector/worker.rs
```diff
@@ -0,0 +1,269 @@
...
+const SQLITE_DATABASE: Option<&CStr> = Some(c"main");
+const SQLITE_TABLE: &CStr = c"collector";
+
+pub static COLLECTOR: Mutex<Option<Connection>> = Mutex::new(None);
+
+pub struct QueryCollector {}
+
+impl QueryCollector {
+    pub fn init() {
```
usamoi: All operations must be deferred until the session starts, so setting up an `init` is inappropriate.

src/collector/worker.rs
```diff
@@ -0,0 +1,269 @@
...
+            Err(e) => {
+                pgrx::warning!("Collector: Error initializing database: {}", e);
+                None
+            }
+        };
+    }
+    pub fn push(query: Query, max_length: u32) {
+        let maintain_statement = "DELETE FROM collector
+            WHERE rowid NOT IN (SELECT rowid FROM collector WHERE database_oid = ?1 AND index_oid = ?2
+            ORDER BY create_at DESC LIMIT ?3) AND database_oid = ?1 AND index_oid = ?2";
```
usamoi: What if users want to keep `1,000,000` rows? I'm suspicious of the efficiency here.

src/collector/worker.rs
```diff
@@ -0,0 +1,269 @@
...
+        let mut lock = match COLLECTOR.try_lock() {
+            Ok(lock) => lock,
+            Err(e) => {
+                pgrx::warning!(
+                    "Collector: Failed to acquire lock on collector connection: {}",
+                    e
+                );
+                return;
+            }
+        };
```
usamoi: It doesn't work.

src/collector/types.rs
```diff
@@ -0,0 +1,283 @@
...
+    );
+}
+
+#[derive(Debug)]
+pub struct DefaultSender {
+    pub send_prob: Option<f64>,
+    pub max_records: u32,
+    pub database_oid: u32,
+    pub table_oid: u32,
+    pub index_oid: u32,
```
usamoi:
```suggestion
    pub index: u32,
```

src/collector/types.rs
```diff
@@ -0,0 +1,283 @@
...
+    fn send_vchordrq(
+        &self,
+        opfamily: crate::index::vchordrq::opclass::Opfamily,
+        vector: vchordrq::types::OwnedVector,
+    );
+    fn send_vchordg(
+        &self,
+        opfamily: crate::index::vchordg::opclass::Opfamily,
+        vector: vchordg::types::OwnedVector,
+    );
```
usamoi:
```suggestion
    fn record(&self, sample: &str);
```

src/collector/types.rs
```diff
@@ -0,0 +1,283 @@
...
+                    self.table_oid,
+                    self.index_oid,
+                    operator,
+                    vector,
+                );
+                if let Some(q) = query {
+                    QueryCollector::push(q, self.max_records);
+                }
+            }
+        }
```
usamoi:
```suggestion
    fn send(
        &self,
        sample: &str,
    ) {
        if let Some(rate) = self.send_prob {
            let mut rng = rand::rng();
            if rng.random_bool(rate) {
                push(sample, self.max_records);
            }
        }
```

src/collector/types.rs
```diff
@@ -0,0 +1,283 @@
...
+    }
```
usamoi:
```suggestion
```

src/collector/types.rs
```diff
@@ -0,0 +1,283 @@
...
+        QueryDump {
+            namespace,
+            table_name,
+            column_name,
+            operator,
+            vector_text,
+            simplified_query,
+        }
+    }
+}
```
usamoi:
```suggestion
```

Do it in SQL.

src/collector/worker.rs
cutecutecat: 1. LGTM. This can make collector more structural.
2. As we collect queries for every index instead of table, I prefer to use a sqlite table for each PostgreSQL index, it can simplify load, maintain and delete.
3. Yes, I agree that it might be lost on backup/restore. Though, as statistical information is often considered volatile, and data that is too outdated holds little value, it's might be acceptable. User can collect it again after restore.

src/collector/worker.rs
cutecutecat: `create_at` is used to eliminate the oldest record. As we perform rolling deletion, RowID may overflow.

src/collector/worker.rs
cutecutecat: It is a so common name, if other extension choose it to store their data, things might be out of control. Maybe `pg_vchord_sample`?

src/collector/types.rs
```diff
@@ -0,0 +1,168 @@
...
+            locktag_type: pgrx::pg_sys::LockTagType::LOCKTAG_ADVISORY as u8,
+            locktag_lockmethodid: pgrx::pg_sys::USER_LOCKMETHOD as u8,
+            locktag_field1: VCHORD_MAGIC,
+            locktag_field2: lock_id,
+            locktag_field3: 0,
+            locktag_field4: 0,
+        };
+        let status = unsafe {
+            pgrx::pg_sys::LockAcquire(
+                &lock_tag as *const _ as *mut _,
```
usamoi: Why not `&raw mut lock_tag`?

src/collector/worker.rs
usamoi: > 3\. Yes, I agree that it might be lost on backup/restore. Though, as statistical information is often considered volatile, and data that is too outdated holds little value, it's might be acceptable. User can collect it again after restore.

If you have actually tried it, the problem is not about the data is volatile. The problem is the data is sync-ed unexpectedly to replicas. Please review the review before replying and don't mark converations resolved before it's resolved.

src/collector/types.rs
```diff
@@ -0,0 +1,168 @@
...
+                    .into_vec()
+                    .iter()
+                    .map(|f| format!("{:.4}", f.to_f32()))
+                    .collect();
+                let joined_elements = vector_format.join(", ");
+                write!(f, "'[{joined_elements}]'")
+            }
+        }
+    }
+}
```
usamoi:
```suggestion
```

I believe I have already stated in https://github.com/tensorchord/VectorChord/pull/321#discussion_r2301120153: use strings.

1. All PostgreSQL types have its `text_in` and `text_out`, so it's general.
2. There is no need to add these code. We don't get benefits from it.

src/collector/types.rs
```diff
@@ -0,0 +1,168 @@
...
+    fn send(&self, sample: &str) {
+        if !self.enable {
+            return;
+        }
+        if let Some(rate) = self.send_prob {
+            let mut rng = rand::rng();
+            if rng.random_bool(rate)
+                && let Err(e) = push(self.database_oid, self.index_oid, sample, self.max_records)
+            {
+                pgrx::warning!("Collector: Error pushing sample: {}", e);
```
usamoi: This warning, if needed, should be placed in the code that handles the SQLite connection.

src/collector/worker.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+    match fs::remove_dir_all(COLLECTOR_DIR) {
+        Ok(_) => (),
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => (),
+        Err(e) => return Err(e.into()),
+    };
+    fs::create_dir_all(COLLECTOR_DIR)?;
+    fs::write(path, format!("{COLLECTOR_VERSION}"))?;
+    Ok(true)
+}
```
usamoi: https://github.com/tensorchord/VectorChord/pull/321#discussion_r2301088850

src/collector/worker.rs
```diff
@@ -0,0 +1,139 @@
...
+// vectorchord-inquiry@tensorchord.ai
+//
+// Copyright (c) 2025 TensorChord Inc.
+use crate::collector::types::{PGLockGuard, Result};
+use std::cell::OnceCell;
+use std::ffi::{CStr, CString};
+use std::fs;
+use std::path::Path;
+
+const SQLITE_DATABASE: Option<&CStr> = Some(c"main");
```
usamoi: It's only used once. Don't split it up.

src/collector/worker.rs
```diff
@@ -0,0 +1,139 @@
...
+    fs::create_dir_all(COLLECTOR_DIR)?;
+    fs::write(path, format!("{COLLECTOR_VERSION}"))?;
+    Ok(true)
+}
+
+fn sqlite_connect(
+    create: bool,
+    database_oid: u32,
+    index_oid: u32,
+) -> Result<Option<rusqlite::Connection>> {
```
usamoi: Why `sqlite_connect` is associated with `index_oid`? Why not keep the connection alive? Why put the logic here, not where the samples are written?

src/collector/worker.rs
```diff
@@ -0,0 +1,139 @@
...
+    let multi_access_lock = PGLockGuard::new(MULTI_ACCESS_LOCK, false);
+    if !multi_access_lock.is_success() {
+        return Ok(());
+    }
+    let mut conn = match sqlite_connect(true, database_oid, index_oid)? {
+        Some(c) => c,
+        None => return Ok(()),
+    };
+    let maintain_statement = format!("DELETE FROM index_{index_oid}
+                WHERE rowid NOT IN (SELECT rowid FROM index_{index_oid} ORDER BY create_at DESC LIMIT ?1)");
```
usamoi: You didn't answer my question in https://github.com/tensorchord/VectorChord/pull/321#discussion_r2301101947.

src/index/gucs.rs
```diff
@@ -27,6 +27,12 @@ pub enum PostgresIo {
     ReadStream,
 }
 
+static VCHORDRQ_LOG_QUERIES_ENABLE: GucSetting<bool> = GucSetting::<bool>::new(false);
```
usamoi: I asked AI, and it didn't properly understand what's `LOG_QUERIES`.

src/index/vchordg/scanners/default.rs
```diff
@@ -104,6 +106,7 @@ impl SearchBuilder for DefaultBuilder {
         let Some(vector) = vector else {
             return Box::new(std::iter::empty()) as Box<dyn Iterator<Item = (f32, [u16; 3], bool)>>;
         };
+        let vector_send = SendVector::Vchordg(vector.clone());
```
usamoi:
```suggestion
```

src/index/vchordrq/scanners/maxsim.rs
```diff
@@ -72,6 +73,7 @@ impl SearchBuilder for MaxsimBuilder {
         options: SearchOptions,
         mut fetcher: impl Fetcher + 'b,
         bump: &'b impl Bump,
+        _sender: impl CollectorSender,
```
usamoi: I think you could try to support this here, respecting PostgreSQL's text representation of `vector[]`.

src/sql/finalize.sql
```diff
@@ -142,6 +142,102 @@ IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchord_vector_
 CREATE FUNCTION quantize_to_scalar8(halfvec) RETURNS scalar8
 IMMUTABLE STRICT PARALLEL SAFE LANGUAGE c AS 'MODULE_PATHNAME', '_vchord_halfvec_quantize_to_scalar8_wrapper';
 
+CREATE FUNCTION _vchordrq_logged_samples(regclass)
```
```suggestion
CREATE FUNCTION vchordrq_logged_samples(regclass)
```

usamoi:
`tests/vchordrq/recall.slt` fails with a probability.

src/collector/worker.rs
cutecutecat: I has added an index at created_at with changed to a range filter delete.

src/index/vchordrq/scanners/maxsim.rs
cutecutecat: The most widely used index is single-vector vchordrq. We can put the maxsim support and vchordg support in phase 2 implement if the record function of them is asked by some user.

src/index/vchordrq/scanners/default.rs
```diff
@@ -108,6 +110,7 @@ impl SearchBuilder for DefaultBuilder {
         let Some(vector) = vector else {
             return Box::new(std::iter::empty()) as Box<dyn Iterator<Item = (f32, [u16; 3], bool)>>;
         };
+        let vector_send = vector.clone();
```
usamoi:
```suggestion
```

src/collector/types.rs
```diff
@@ -0,0 +1,136 @@
...
+                    pgrx::pg_sys::ExclusiveLock as _,
+                    false,
+                );
+            }
+        }
+    }
+}
+
+pub trait CollectorSender {
+    fn is_enable(&self) -> bool;
```
usamoi:
```suggestion
    fn is_enabled(&self) -> bool;
```

src/index/vchordrq/scanners/default.rs
```diff
@@ -689,6 +692,18 @@ impl SearchBuilder for DefaultBuilder {
...
+                OwnedVector::Vecf32(v) => v.into_vec().iter().map(|f| format!("{f}")).collect(),
+                OwnedVector::Vecf16(v) => v
+                    .into_vec()
+                    .iter()
+                    .map(|f| format!("{}", f.to_f32()))
+                    .collect(),
+            };
+            let vector_text = format!("[{}]", vector_text.join(","));
+            sender.send(&vector_text);
+        }
```
usamoi:
```suggestion
        if sender.is_enabled() {
            match &vector {
                OwnedVector::Vecf32(v) => {
                    sender.send(&vector_out(v.as_borrowed()));
                },
                OwnedVector::Vecf16(v) => {
                    sender.send(&halfvec_out(v.as_borrowed()));                    
                }
            }
        }

        // ./src/collector/text.rs

        fn vector_out(vector: VectBorrowed<'_, f32>) -> String {
            let mut result = String::from("[");
            for x in vector.slice() {
                if result.ends_with('[') {
                    result.push(', ');
                }
                result.push_str(&x.to_string());
            }
            result.push(']');
            result
        }
        
        fn halfvec_out(vector: VectBorrowed<'_, f16>) -> String {
            let mut result = String::from("[");
            for x in vector.slice() {
                if result.ends_with('[') {
                    result.push(', ');
                }
                result.push_str(&x.to_string());
            }
            result.push(']');
            result
        }
```

src/collector/types.rs
```diff
@@ -0,0 +1,136 @@
...
+            if self.success {
+                pgrx::pg_sys::LockRelease(
+                    &raw mut self.lock_tag,
+                    pgrx::pg_sys::ExclusiveLock as _,
+                    false,
+                );
+            }
+        }
+    }
+}
```
usamoi:
```suggestion
```

src/collector/types.rs
```diff
@@ -0,0 +1,136 @@
...
+
+#[derive(Debug)]
+pub struct DefaultSender {
+    pub enable: bool,
+    pub send_prob: Option<f64>,
+    pub max_records: u32,
+    pub database_oid: u32,
+    pub index_oid: u32,
+}
```
usamoi:
```suggestion
#[derive(Debug)]
pub struct DefaultSender {
    pub enable: bool,
    pub rate: Option<f64>,
    pub max_records: u32,
    pub index: u32,
}
```

src/collector/worker.rs
```diff
@@ -0,0 +1,136 @@
...
+
+const COLLECTOR_DIR: &str = "pg_vchord_sample";
+const COLLECTOR_VERSION: u32 = 1;
+
+const MULTI_ACCESS_LOCK: u32 = 0;
+
+thread_local! {
+    static PATH_READY: OnceCell<bool> = const { OnceCell::new() };
+    static DB_CONNECTION: RefCell<HashMap<u32, rusqlite::Connection>> = RefCell::new(HashMap::new());
+}
```
usamoi:
```suggestion
```

src/collector/worker.rs
```diff
@@ -0,0 +1,136 @@
...
+use std::cell::{OnceCell, RefCell};
+use std::collections::HashMap;
+use std::ffi::CString;
+use std::fs;
+use std::path::Path;
+
+const COLLECTOR_DIR: &str = "pg_vchord_sample";
+const COLLECTOR_VERSION: u32 = 1;
+
+const MULTI_ACCESS_LOCK: u32 = 0;
```
usamoi:
```suggestion
```

src/collector/worker.rs
cutecutecat: Sorry for my misunderstand.

The query collector is tolerable to have more or less records, if each record are corresponding to the correct index oid. 

Suppose pgdata samples is unexpected syned to the replica. If the index oid / table oid is also synced correctly: Everything is okay, the only effect is that we will get additional records from elsewhere, but they are also belongs to "records of that index". So I think this is maybe also acceptable.

src/collector/worker.rs
```diff
@@ -0,0 +1,136 @@
...
+        let mut rows = stmt.query(())?;
+        let mut result = Vec::new();
+        while let Some(row) = rows.next().unwrap_or(None) {
+            let sample = row.get::<usize, String>(0)?;
+            result.push(sample);
+        }
+        Ok(result)
+    })
+    .map(|opt| opt.unwrap_or_default())
+}
```
usamoi:
```suggestion
struct Collector {
    connection: rusqlite::Connection,
}

static COLLECTOR: PgRefCell<Option<Collector>> = PgRefCell::<Option<Collector>>::new(None);

pub fn get() -> Option<PgRefMut<Collector>> {
    // checks if it's postmaster, if yes, return `None`
    // checks if there is a vaild database oid, if yes, return `None`
    // checks if there is a value in `COLLECTOR.get()`, if yes, return it
    // open the connection
}

impl Collector {
    pub fn push(&mut self, index: u32, sample: &str, max_records: u32) {
        // start a transaction
        // checks the version, if it's outdated, remove the table
        // create the table if it doesn't exist
        // insert the record into the table
        // remove the outdated records
        // end a transaction
    }
    
    pub fn dump(&mut self, index: u32) -> Vec<String> {
        // start a transaction
        // checks the version, if it's outdated, remove the table
        // create the table if it doesn't exist
        // dump everything from the table
        // end a transaction
    }

    pub fn delete(&mut self, index: u32) -> Vec<String> {
        // start a transaction
        // delete the table if it exists
        // end a transaction
    }
}
```

src/collector/worker.rs
usamoi: > Everything is okay, the only effect is that we will get additional records from elsewhere, but they are also belongs to "records of that index".

It may break sqlite database file and it's unacceptable.

src/collector/worker.rs
usamoi: Here we needs something like `PgRefCell`. It protects the data by checking if it's main thread, instead of performing atomic operations.

usamoi: Here, the use of PostgreSQL locking mechanism is problematic, so I suggest not adding this extra lock for now.

usamoi (changes requested):

src/index/vchordrq/scanners/default.rs
cutecutecat: Here the vector is moved in:
```
let unprojected = if let OwnedVector::Vecf32(vector) = vector {
```

usamoi: > Here the vector is moved in:
> 
> ```
> let unprojected = if let OwnedVector::Vecf32(vector) = vector {
> ```

Clone it here.

cutecutecat:
## Query record latency test

| Latency                             | dim  | minimum | maximum | majority |
| ----------------------------------- | ---- | ------- | ------- | -------- |
| insert at <1000 records + maintain  | 3    | 3ms     | 5ms     | 5ms      |
| insert at ~5000 records + maintain  | 3    | 5ms     | 6ms     | 7ms      |
| insert at ~10000 records + maintain | 3    | 10ms    | 13ms    | 10ms     |
| insert at <1000 records + maintain  | 768  | 3ms     | 6ms     | 5ms      |
| insert at ~5000 records + maintain  | 768  | 4ms     | 6ms     | 5ms      |

- Set maximum of `query_sampling_max_records` to 5000.

usamoi:
> Query record latency test

What does this show? Shouldn't we conduct a controlled experiment with variables?

src/collector/types.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub enable: bool,
+    pub rate: Option<f64>,
+    pub max_records: u32,
+    pub index: u32,
+}
+
+pub struct PgRefCell<T>(std::cell::RefCell<T>);
+
+unsafe impl<T: Send> Send for PgRefCell<T> {}
+unsafe impl<T> Sync for PgRefCell<T> {}
```
usamoi:
```suggestion
unsafe impl<T: Send> Sync for PgRefCell<T> {}
```

src/collector/types.rs
```diff
@@ -0,0 +1,58 @@
...
+
+impl<T> PgRefCell<T> {
+    pub const unsafe fn new(x: T) -> Self {
+        Self(std::cell::RefCell::new(x))
+    }
+    pub fn borrow_mut(&self) -> RefMut<'_, T> {
+        self.0.borrow_mut()
+    }
```
usamoi:
```suggestion
    pub fn borrow_mut(&self) -> RefMut<'_, T> {
        assert!(crate::is_main(), "cannot borrow the value outside main thread");
        self.0.borrow_mut()
    }
```

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+// regarding the licenses, please contact us at:
+// vectorchord-inquiry@tensorchord.ai
+//
+// Copyright (c) 2025 TensorChord Inc.
+use crate::collector::types::PgRefCell;
+use std::cell::RefMut;
+use std::fs;
+use std::path::Path;
+
+const COLLECTOR_DIR: &str = "pg_vchord_sample";
```
usamoi:
```suggestion
const COLLECTOR_DIR: &str = "pg_vchord_sampling";
```

src/collector/types.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub index: u32,
+}
+
+pub struct PgRefCell<T>(std::cell::RefCell<T>);
+
+unsafe impl<T: Send> Send for PgRefCell<T> {}
+unsafe impl<T> Sync for PgRefCell<T> {}
+
+impl<T> PgRefCell<T> {
+    pub const unsafe fn new(x: T) -> Self {
```
usamoi:
```suggestion
    pub const fn new(x: T) -> Self {
```

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+const COLLECTOR_VERSION: u32 = 1;
+
+static CONNECTION: PgRefCell<Option<rusqlite::Connection>> =
+    unsafe { PgRefCell::<Option<rusqlite::Connection>>::new(None) };
+
+fn get<'a>() -> Option<RefMut<'a, rusqlite::Connection>> {
+    if unsafe { !pgrx::pg_sys::IsUnderPostmaster } {
```
usamoi:
```c
/*
 * IsPostmasterEnvironment is true in a postmaster process and any postmaster
 * child process; it is false in a standalone process (bootstrap or
 * standalone backend).  IsUnderPostmaster is true in postmaster child
 * processes.  Note that "child process" includes all children, not only
 * regular backends.  These should be set correctly as early as possible
 * in the execution of a process, so that error handling will do the right
 * things if an error should occur during process initialization.
 *
 * These are initialized for the bootstrap/standalone case.
 */
```

I think this helps to determine whether this is a running backend.

```c
/*
 * IsBackendPid -- is a given pid a running backend
 *
 * This is not called by the backend, but is called by external modules.
 */
bool
IsBackendPid(int pid)
{
	return (BackendPidGetProc(pid) != NULL);
}
```

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+    let database_oid = unsafe { pgrx::pg_sys::MyDatabaseId.to_u32() };
+    if database_oid == 0 {
+        return None;
+    }
+    let mut connection = CONNECTION.borrow_mut();
+    if connection.is_none() {
+        if !Path::new(COLLECTOR_DIR).exists() {
+            fs::create_dir_all(COLLECTOR_DIR).ok()?;
+        }
+        let p = format!("pg_vchord_sample/database_{database_oid}.sqlite");
```
usamoi: `COLLECTOR_DIR` is used sometimes, and not used sometimes.

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+            rusqlite::Connection::open(&p).ok()?
+        } else {
+            conn
+        };
+        // SET journal_mode=WAL for read/write concurrency
+        conn.pragma_update(Some("main"), "journal_mode", "WAL")
+            .ok()?;
+        conn.pragma_update(Some("main"), "user_version", COLLECTOR_VERSION)
+            .ok()?;
+        connection.replace(conn);
```
usamoi:
```suggestion
        connection.insert(conn);
```

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+    if connection.is_none() {
+        return None;
+    }
+    Some(RefMut::map(connection, |c| c.as_mut().unwrap()))
```
usamoi:
```suggestion
    RefMut::filter_map(connection, |c| c.as_mut()).ok()
```

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+            fs::create_dir_all(COLLECTOR_DIR).ok()?;
+        }
+        let p = format!("pg_vchord_sample/database_{database_oid}.sqlite");
+        let conn = rusqlite::Connection::open(&p).ok()?;
+        let version: u32 = conn
+            .pragma_query_value(Some("schema"), "user_version", |row| row.get(0))
+            .unwrap_or(COLLECTOR_VERSION);
+        let conn = if version != COLLECTOR_VERSION && version != 0 {
+            drop(conn);
+            fs::remove_file(&p).ok()?;
```
usamoi: race condition

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+fn get<'a>() -> Option<RefMut<'a, rusqlite::Connection>> {
+    if unsafe { !pgrx::pg_sys::IsUnderPostmaster } {
+        return None;
+    }
+    let database_oid = unsafe { pgrx::pg_sys::MyDatabaseId.to_u32() };
+    if database_oid == 0 {
+        return None;
+    }
+    let mut connection = CONNECTION.borrow_mut();
+    if connection.is_none() {
```
usamoi: You can extract the code under this branch into an immediately invoked closure, assign the returned `Ok` to connection, and ignore the returned `Err` or print logs.

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+            .ok()?;
+        connection.replace(conn);
+    }
+    if connection.is_none() {
+        return None;
+    }
+    Some(RefMut::map(connection, |c| c.as_mut().unwrap()))
+}
+
+pub fn push(index_oid: u32, sample: &str, max_records: u32) {
```
usamoi:
```suggestion
pub fn push(index: u32, sample: &str, max_records: u32) {
```

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+    let mut connection = match get() {
+        Some(c) => c,
+        None => return,
+    };
+    let init_statement = format!(
+        "
+        CREATE TABLE IF NOT EXISTS index_{index_oid} (id INTEGER PRIMARY KEY, sample TEXT, create_at REAL);
+        CREATE INDEX IF NOT EXISTS i ON index_{index_oid} (create_at DESC, id DESC);
```
usamoi: Why not just use u128 as primary key?

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+        "
+    );
+    let insert_statement = format!(
+        "INSERT INTO index_{index_oid} (sample, create_at) VALUES (?1, unixepoch('subsec'))"
+    );
+    let maintain_statement = format!(
+        "DELETE FROM index_{index_oid} WHERE (create_at, id) < (
+            SELECT create_at, id FROM index_{index_oid}
+            ORDER BY create_at DESC, id DESC
+            LIMIT 1 OFFSET ({max_records} - 1))"
```
usamoi: I now think that as long as I delete the oldest record and add the newest one, that would be sufficient here.

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+    };
+    if let Err(e) = tx.execute(&insert_statement, (sample,)) {
+        pgrx::debug1!("Collector: Error inserting sample: {}", e);
+        return;
+    }
+    if let Err(e) = tx.execute(&maintain_statement, ()) {
+        pgrx::debug1!("Collector: Error maintaining samples: {}", e);
+        return;
+    }
+    let _ = tx.commit();
```
usamoi: Extract the code into an immediately invoked closure, and ignore the returned `Err` or print logs.

src/collector/worker.rs
```diff
@@ -0,0 +1,150 @@
...
+            return Vec::new();
+        }
+    };
+    let mut rows = match stmt.query(()) {
+        Ok(r) => r,
+        Err(e) => {
+            pgrx::debug1!("Collector: Error querying samples: {}", e);
+            return Vec::new();
+        }
+    };
```
usamoi: Extract the code into an immediately invoked closure, and ignore the returned `Err` or print logs.

src/index/functions.rs
```diff
@@ -84,3 +86,26 @@ impl Drop for Index {
...
+    };
+    if pg_class.relkind() != PgClassRelkind::Index {
+        pgrx::error!("the relation {:?} is not an index", pg_class.relname());
+    }
+    if pg_class.relam() != pg_am.oid() {
+        pgrx::error!("the index {:?} is not a vchordrq index", pg_class.relname());
+    }
+    // The user must have access to the index, if not, raise an error from Postgres.
+    let _relation = Index::open(indexrelid, pgrx::pg_sys::AccessShareLock as _);
+    let max_records = crate::index::gucs::vchordrq_query_sampling_max_records();
```
usamoi: We should return all rows here. `max_records` will control the output too strangely.

src/collector/worker.rs
cutecutecat: Renamed to `pgsql_tmp_vchord_sampling` due to discussions.

Background: The SQLite file will be unstable if it is copied by pg_basebackup while it is open and being appended to. To avoid this, the directory name should start with `pgsql_tmp`, then it will be excluded by pg_basebackup.

ref:
https://www.postgresql.org/docs/current/app-pgbasebackup.html
https://github.com/postgres/postgres/blob/aba8f61c30911cbac1a310b7d21777e04711c66e/src/backend/backup/basebackup.c#L1263
https://github.com/tensorchord/VectorChord/pull/321#discussion_r2314782353

cutecutecat: Verified. The exclude strategy worked.

src/collector/worker.rs
```diff
@@ -0,0 +1,142 @@
...
+    let init_statement = format!(
+        "
+        CREATE TABLE IF NOT EXISTS index_{index} (sample TEXT, create_at REAL);
+        CREATE INDEX IF NOT EXISTS i ON index_{index} (create_at);
+        "
+    );
+    let insert_statement =
+        format!("INSERT INTO index_{index} (sample, create_at) VALUES (?1, unixepoch('subsec'))");
+    let count_statement = format!("SELECT COUNT(create_at) FROM index_{index}");
```
usamoi: This should not be performed. This will only result in what was originally a full table scan once now becoming a full table scan twice.

src/lib.rs
```diff
@@ -16,6 +16,7 @@
 #![allow(unused_crate_dependencies)]
 #![warn(ffi_unwind_calls)]
 
+mod collector;
```
usamoi: Rename `collector` to `recorder`.
Rename `CollectorSender` to `Recorder`.
Rename `sender` to `recorder`.

PR #325, opened 2025-08-28
feat: enable fast math in simd

usamoi:


VoVAllen:
do you see any performance improvements?

PR #334, opened 2025-09-09
chord: add index_name into recorder

cutecutecat:


src/recorder/text.rs
```diff
@@ -38,3 +38,23 @@ pub fn halfvec_out(vector: VectBorrowed<'_, f16>) -> String {
...
+    if elements.is_empty() {
+        return false;
+    }
+    for element in elements {
+        if element.is_empty() || element.parse::<f32>().is_err() {
+            return false;
+        }
+    }
+    true
+}
```
usamoi: I think we can just guide users to use `quote_literal` and `quote_ident` in SQL.

src/sql/finalize.sql
```diff
@@ -151,8 +151,9 @@ RETURNS TABLE(
     schema_name NAME,
     table_name NAME,
     column_name NAME,
+    index_name NAME,
```
```
    schema_name NAME,
    index_name NAME,
    table_name NAME,
    column_name NAME,
```

is better to me.

PR #347, opened 2025-09-19
perf: reduce lock contention

usamoi:
```sql
create index on laion1m768_train using vchordrq (embedding vector_ip_ops) with (options = $$
build.pin = 2
build.internal.lists = [256, 4096]
build.internal.build_threads = 20
build.internal.spherical_centroids = true
degree_of_parallelism = 40
$$);
```

`degree_of_parallelism` is 32 by default. It reduces lock contention during insertion, especially, in a build where `lists.len() != 1`.

`build.pin` is `-1` by default. `build.pin = 2` caches almost everything in memory, which speeds build but spends more memory. `build.pin = false` and `build.pin = true` still works, and they're mapped to `build.pin = -1` and `build.pin = 1`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,9 +97,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.99"
+version = "1.0.100"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b0674a1ddeecb70197781e945de4b3b8ffb61fa939a5597bcf48503737663100"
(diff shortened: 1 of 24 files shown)


VoVAllen:
What is vector_tapes? It should be a hidden config named like `_vector_tapes`

usamoi:
> What is vector_tapes? It should be a hidden config named like `_vector_tapes`

It's an experimental trick to speed parallel build, and won't be merged as-is.

cutecutecat:
With `vector_tapes = 32` and 16 workers, the assignment of `LAION-100M` can be accerated from 3300s to 1800s.

PR #350, opened 2025-09-26
chore: upload 0.5.3 schema scripts

usamoi:


VoVAllen:
Is https://github.com/tensorchord/VectorChord/pull/347 a breaking change?

usamoi:
> Is #347 a breaking change?

Yes, so that won't be in `0.5.3`.

PR #354, opened 2025-10-07
Update scripts for ease of use

mschoema:
I tried using the scripts to generate centroids from a dataset and I hit a few issues, so I though I would send a PR with some improvements. The specific changes are detailed in the commit messages. Let me know if you are interested in these changes. I can also revert individual commits if you do not want these specific changes.


mschoema:

PR #358, opened 2025-10-15
feat: approximate kmeans and dimension reduction 

cutecutecat:
## Background
This PR provides a much more faster kmeans algorithm at the cost of less accuracy.

```mermaid
graph TD
0[vectors] -- no reduction --> 1[samples]
0[vectors] -- reduction --> 1[samples]
1[samples] -- last layer --> 2["appro_kmeans(n)"]
2["appro_kmeans(n)"] --> 3["get m = sqrt(n"]
3["get m = sqrt(n)"] --> 4["KMeans(m)"]
4["KMeans(m)"] --> 5[assign samples to m clusters]
5[assign samples to m clusters] --> 6[calculate sub-clusters size by allocation]
6[calculate sub-clusters size by allocation] --> 7[KMeans for all n clusters]
7[KMeans for all n clusters] -- reduction --> 8[get assignment of n clusters]
8[get assignment of n clusters] --> 9[retravel heap and get new centroids]
7[KMeans for all n clusters] -- no reduction --> 10[centroids]
9[retravel heap and get new centroids] --> 12[recompute centroids from assignment]
12[recompute centroids from assignment] --> 10[centroids]
```

## Usage
```sql
CREATE INDEX ON laion USING vchordrq (embedding vector_ip_ops) WITH (options = $$
residual_quantization = false
build.pin = true
[build.internal]
lists = [160000]
build_threads = 16
spherical_centroids = true
kmeans_algorithm = { hierarchical = {} }
$$);

CREATE INDEX ON laion USING vchordrq (embedding vector_ip_ops) WITH (options = $$
residual_quantization = false
build.pin = true
[build.internal]
lists = [400, 160000]
build_threads = 16
spherical_centroids = true
kmeans_dimension_reduction = 2 # optional dim: dimension reduction to reduce memory usage
kmeans_algorithm = { hierarchical = {} }
$$);
# Build time for this entry: 20m for LAION-100M on 16c
```

img_v3_02rm_4181dccb-0003-450c-9f15-8ec5fc3756bg

### DIFF
--- a/crates/k_means/src/lib.rs
+++ b/crates/k_means/src/lib.rs
@@ -13,18 +13,18 @@
 // Copyright (c) 2025 TensorChord Inc.
 
 pub mod flat;
+pub mod hierarchical;
 pub mod quick;
 pub mod rabitq;
 pub mod square;
 
-use crate::square::Square;
+use crate::square::{Square, SquareMut};
 use rand::rngs::StdRng;
 
 pub struct This {
     pool: rayon::ThreadPool,
     rng: StdRng,
     d: usize,
-    samples: Square,
     c: usize,
     centroids: Square,
     targets: Vec<usize>,
@@ -45,16 +45,17 @@ pub trait KMeans {
             });
         });
     }
+    fn index(&mut self) -> Box<dyn Fn(&[f32]) -> u32 + '_>;
     fn finish(self: Box<Self>) -> Square;
 }
 
-pub fn k_means(
+pub fn k_means<'a>(
     d: usize,
-    samples: Square,
+    samples: SquareMut<'a>,
     c: usize,
     num_threads: usize,
     seed: [u8; 32],
-) -> Box<dyn KMeans> {
+) -> Box<dyn KMeans + 'a> {
     assert!(d > 0 && c > 0 && num_threads > 0);
     let n = samples.len();
     if n <= c {
@@ -64,4 +65,16 @@ pub fn k_means(
     } else {
         rabitq::new(d, samples, c, num_threads, seed)
     }
+}
+
+pub fn hierarchical_k_means<'a>(
+    d: usize,
+    samples: SquareMut<'a>,
+    c: usize,
+    num_threads: usize,
+    seed: [u8; 32],
+    is_spherical: bool,
+) -> Box<dyn KMeans + 'a> {
+    assert!(d > 0 && c > 0 && num_threads > 0);
+    hierarchical::new(d, samples, c, num_threads, seed, is_spherical)
 }
--- a/crates/k_means/src/square.rs
+++ b/crates/k_means/src/square.rs
@@ -59,6 +59,13 @@ impl Square {
     pub fn truncate(&mut self, len: usize) {
         self.p.truncate(self.d * len);
     }
+    #[inline]
+    pub fn as_mut_view(&mut self) -> SquareMut<'_> {
+        SquareMut {
+            d: self.d,
+            p: self.p.as_mut_slice(),
+        }
+    }
 }
 
 impl std::ops::Index<usize> for Square {
@@ -112,5 +119,85 @@ impl<'a> rayon::prelude::IntoParallelIterator for &'a mut Square {
 
     fn into_par_iter(self) -> Self::Iter {
         rayon::slice::ParallelSliceMut::par_chunks_exact_mut(self.p.as_mut_slice(), self.d)
+    }
+}
+
+#[derive(Debug)]
+pub struct SquareMut<'a> {
+    d: usize,
+    p: &'a mut [f32],
+}
+impl<'a> SquareMut<'a> {
+    #[inline]
+    pub fn d(&self) -> usize {
+        self.d
+    }
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.p.len() / self.d
+    }
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.p.is_empty()
+    }
+    pub fn new(d: usize, p: &'a mut [f32]) -> Self {
+        Self { d, p }
+    }
+    #[inline]
+    pub fn iter_mut<'b>(&'b mut self) -> std::slice::ChunksExactMut<'b, f32>
+    where
+        'a: 'b,
+    {
+        self.p.chunks_exact_mut(self.d)
+    }
+    #[inline]
+    pub fn par_iter_mut<'b>(&'b mut self) -> rayon::slice::ChunksExactMut<'b, f32>
+    where
+        'a: 'b,
+    {
+        rayon::slice::ParallelSliceMut::par_chunks_exact_mut(self.p, self.d)
+    }
+    #[inline]
+    pub fn row(&self, i: usize) -> &[f32] {
+        let d = self.d;
+        &self.p[i * d..(i + 1) * d]
+    }
+    #[inline]
+    pub fn row_mut(&mut self, i: usize) -> &mut [f32] {
+        let d = self.d;
+        &mut self.p[i * d..(i + 1) * d]
+    }
+    pub fn copy_within<R: std::ops::RangeBounds<usize>>(&mut self, src: R, dest: usize) {
+        let src_start = src.start_bound().map(|x| self.d * x);
+        let src_end = src.end_bound().map(|x| self.d * x);
+        self.p.copy_within((src_start, src_end), self.d * dest);
+    }
+    #[inline]
+    pub fn into_inner(self) -> (usize, &'a mut [f32]) {
+        (self.d, self.p)
+    }
+}
+
+impl std::ops::Index<usize> for SquareMut<'_> {
+    type Output = [f32];
+
+    fn index(&self, index: usize) -> &Self::Output {
+        &self.p[self.d * index..][..self.d]
+    }
+}
+
+impl std::ops::IndexMut<usize> for SquareMut<'_> {
+    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
+        &mut self.p[self.d * index..][..self.d]
+    }
+}
+
+impl<'a> rayon::prelude::IntoParallelIterator for &'a mut SquareMut<'a> {
+    type Item = &'a mut [f32];
+
+    type Iter = rayon::slice::ChunksExactMut<'a, f32>;
+
+    fn into_par_iter(self) -> Self::Iter {
+        rayon::slice::ParallelSliceMut::par_chunks_exact_mut(self.p, self.d)
     }
 }
--- a/src/index/vchordrq/am/am_build.rs
+++ b/src/index/vchordrq/am/am_build.rs
@@ -25,6 +25,7 @@ use crate::index::vchordrq::types::*;
 use index::relation::{
     Page, PageGuard, Relation, RelationRead, RelationReadTypes, RelationWrite, RelationWriteTypes,
 };
+use k_means::flat::k_means_lookup;
 use k_means::square::Square;
 use simd::Floating;
 use std::ffi::CStr;
@@ -1274,8 +1275,20 @@ fn make_internal_build(
 ) -> Vec<Structure<Normalized>> {
     use humansize::{BINARY, format_size};
     use std::iter::once;
+    let (reduction, sample_dim) = match internal_build.kmeans_dimension_reduction {
+        None => (false, vector_options.dims as usize),
+        Some(d) if d < vector_options.dims => (true, d as usize),
+        Some(d) => {
+            pgrx::warning!(
+                "ignoring `kmeans_dimension_reduction = {}` because it is less than the vector dimension {}",
+                d,
+                vector_options.dims
+            );
+            (false, vector_options.dims as usize)
+        }
+    };
     {
-        let d = vector_options.dims as u64;
+        let d = sample_dim as u64;
         let c = internal_build.lists.last().copied().unwrap_or_default() as u64;
         let f = internal_build.sampling_factor as u64;
         let t = internal_build.build_threads as u64;
@@ -1295,8 +1308,7 @@ fn make_internal_build(
             .last()
             .map(|x| x.saturating_mul(internal_build.sampling_factor))
             .unwrap_or_default();
-        let mut samples =
-            Square::with_capacity(vector_options.dims as _, max_number_of_samples as _);
+        let mut samples = Square::with_capacity(sample_dim, max_number_of_samples as _);
         if samples.len() >= max_number_of_samples as usize {
             break 'a samples;
         }
@@ -1307,7 +1319,7 @@ fn make_internal_build(
                 let vectors = unsafe { opfamily.store(datum) };
                 if let Some(vectors) = vectors {
                     for (vector, _) in vectors {
-                        let x = match vector {
+                        let mut x = match vector {
                             OwnedVector::Vecf32(x) => VectOwned::normalize(x),
                             OwnedVector::Vecf16(x) => VectOwned::normalize(x),
                         };
@@ -1316,9 +1328,13 @@ fn make_internal_build(
                             x.len() as u32,
                             "invalid vector dimensions"
                         );
+                        if reduction {
+                            rabitq::rotate::rotate_inplace(&mut x);
+                            x.truncate(sample_dim);
+                        }
                         samples.push_slice(x.as_slice());
                         if samples.len() >= max_number_of_samples as usize {
-                            break;
+                            break 'a samples;
                         }
                     }
                 }
@@ -1329,7 +1345,7 @@ fn make_internal_build(
     let mut result = Vec::<Structure<Normalized>>::new();
     let mut samples = Some(samples);
     for w in internal_build.lists.iter().rev().copied().chain(once(1)) {
-        let input = if let Some(structure) = result.last() {
+        let mut input = if let Some(structure) = result.last() {
             let mut input =
                 Square::with_capacity(vector_options.dims as _, structure.centroids.len());
             for slice in structure.centroids.iter() {
@@ -1343,7 +1359,7 @@ fn make_internal_build(
         };
         let num_threads = internal_build.build_threads as _;
         let num_points = input.len();
-        let num_dims = vector_options.dims as usize;
+        let num_dims = input.d();
         let num_lists = w as usize;
         let num_iterations = internal_build.kmeans_iterations as _;
         if result.is_empty() {
@@ -1358,7 +1374,23 @@ fn make_internal_build(
                 "clustering: starting, using {num_threads} threads, clustering {num_points} vectors of {num_dims} dimension into {num_lists} clusters, in {num_iterations} iterations"
             );
         }
-        let mut f = k_means::k_means(num_dims, input, num_lists, num_threads, [7; 32]);
+        let mut f = {
+            let view = input.as_mut_view();
+            if result.last().is_none()
+                && let KMeansAlgorithm::Hierarchical {} = internal_build.kmeans_algorithm
+            {
+                k_means::hierarchical_k_means(
+                    num_dims,
+                    view,
+                    num_lists,
+                    num_threads,
+                    [7; 32],
+                    internal_build.spherical_centroids,
+                )
+            } else {
+                k_means::k_means(num_dims, view, num_lists, num_threads, [7; 32])
+            }
+        };
         if internal_build.spherical_centroids {
             f.sphericalize();
         }
@@ -1381,7 +1413,45 @@ fn make_internal_build(
                 f.sphericalize();
             }
         }
-        let centroids = f.finish();
+        let centroids = if reduction && result.last().is_none() {
+            let mut sample = sampler.sample();
+            let mut next_sample = || -> Option<(Vec<f32>, Vec<f32>)> {
+                let mut tuple = sample.next()?;
+                let (values, is_nulls) = tuple.build();
+                let datum = (!is_nulls[0]).then_some(values[0]);
+                if let Some(datum) = datum {
+                    let vectors = unsafe { opfamily.store(datum) };
+                    if let Some(vectors) = vectors {
+                        if let Some((vector, _)) = vectors.into_iter().next() {
+                            let x = match vector {
+                                OwnedVector::Vecf32(x) => VectOwned::normalize(x),
+                                OwnedVector::Vecf16(x) => VectOwned::normalize(x),
+                            };
+                            let mut y = rabitq::rotate::rotate(&x);
+                            y.truncate(sample_dim);
+                            return Some((x, y));
+                        }
+                        None
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
+            };
+            k_means_dimension_restore(
+                vector_options.dims as usize,
+                num_lists,
+                num_points,
+                internal_build.spherical_centroids,
+                [7; 32],
+                f.index(),
+                &mut next_sample,
+            )
+        } else {
+            f.finish()
+        };
+
         if result.is_empty() {
             let percentage = 100;
             let default = BuildPhase::from_code(BuildPhaseCode::InternalBuild);
@@ -1395,17 +1465,6 @@ fn make_internal_build(
         if let Some(structure) = result.last() {
             let mut children = vec![Vec::new(); centroids.len()];
             for i in 0..structure.len() as u32 {
-                pub fn k_means_lookup(vector: &[f32], centroids: &Square) -> usize {
-                    assert_ne!(centroids.len(), 0);
-                    let mut result = (f32::INFINITY, 0);
-                    for i in 0..centroids.len() {
-                        let dis = f32::reduce_sum_of_d2(vector, &centroids[i]);
-                        if dis <= result.0 {
-                            result = (dis, i);
-                        }
-                    }
-                    result.1
-                }
                 let target = k_means_lookup(&structure.centroids[i as usize], &centroids);
                 children[target].push(i);
             }
@@ -1426,6 +1485,45 @@ fn make_internal_build(
         }
     }
     result
+}
+
+fn k_means_dimension_restore<'a>(
+    d: usize,
+    c: usize,
+    num_points: usize,
+    is_spherical: bool,
+    seed: [u8; 32],
+    index: Box<dyn Fn(&[f32]) -> u32 + 'a>,
+    mut next_sample: impl FnMut() -> Option<(Vec<f32>, Vec<f32>)>,
+) -> Square {
+    use rand::rngs::StdRng;
+    use rand::{Rng, SeedableRng};
+    let mut centroids = Square::from_zeros(d, c);
+    let mut count = vec![0_u32; c];
+    let mut traveled = 0;
+    let mut rng = StdRng::from_seed(seed);
+    while let Some((original, reduction)) = next_sample()
+        && traveled < num_points
+    {
+        let cid = index(&reduction) as usize;
+        f32::vector_add_inplace(&mut centroids[cid], &original);
+        count[cid] += 1;
+        traveled += 1;
+    }
+    let mut ret = Square::new(d);
+    for (i, centroid) in centroids.as_mut_view().iter_mut().enumerate() {
+        if count[i] == 0 {
+            centroid.fill_with(|| rng.random_range(-1.0f32..1.0f32));
+        } else {
+            f32::vector_mul_scalar_inplace(centroid, 1.0 / count[i] as f32);
+        }
+        if is_spherical {
+            let l = f32::reduce_sum_of_x2(centroid).sqrt();
+            f32::vector_mul_scalar_inplace(centroid, 1.0 / l);
+        }
+        ret.push_slice(centroid);
+    }
+    ret
 }
 
 #[allow(clippy::collapsible_else_if)]
--- a/src/index/vchordrq/types.rs
+++ b/src/index/vchordrq/types.rs
@@ -27,6 +27,14 @@ impl Default for VchordrqDefaultBuildOptions {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "snake_case")]
+pub enum KMeansAlgorithm {
+    Lloyd {},
+    Hierarchical {},
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, Validate)]
 #[serde(deny_unknown_fields)]
 pub struct VchordrqInternalBuildOptions {
@@ -44,6 +52,11 @@ pub struct VchordrqInternalBuildOptions {
     #[serde(default = "VchordrqInternalBuildOptions::default_build_threads")]
     #[validate(range(min = 1, max = 255))]
     pub build_threads: u16,
+    #[serde(default = "VchordrqInternalBuildOptions::default_kmeans_algorithm")]
+    pub kmeans_algorithm: KMeansAlgorithm,
+    #[serde(default = "VchordrqInternalBuildOptions::default_kmeans_dimension_reduction")]
+    #[validate(range(min = 1, max = 16000))]
+    pub kmeans_dimension_reduction: Option<u32>,
 }
 
 impl VchordrqInternalBuildOptions {
@@ -71,6 +84,12 @@ impl VchordrqInternalBuildOptions {
     fn default_build_threads() -> u16 {
         1
     }
+    fn default_kmeans_algorithm() -> KMeansAlgorithm {
+        KMeansAlgorithm::Lloyd {}
+    }
+    fn default_kmeans_dimension_reduction() -> Option<u32> {
+        None
+    }
 }
 
 impl Default for VchordrqInternalBuildOptions {
@@ -81,6 +100,8 @@ impl Default for VchordrqInternalBuildOptions {
             sampling_factor: Self::default_sampling_factor(),
             kmeans_iterations: Self::default_kmeans_iterations(),
             build_threads: Self::default_build_threads(),
+            kmeans_algorithm: Self::default_kmeans_algorithm(),
+            kmeans_dimension_reduction: Self::default_kmeans_dimension_reduction(),
         }
     }
 }
--- a/tests/vchordrq/approximate.slt
+++ b/tests/vchordrq/approximate.slt
@@ -1,0 +1,57 @@
+statement ok
+CREATE TABLE t (val vector(3));
+
+statement ok
+INSERT INTO t (val) SELECT ARRAY[random(), random(), random()]::real[] FROM generate_series(1, 1000);
+
+statement ok
+CREATE INDEX ON t USING vchordrq (val vector_l2_ops)
+WITH (options = $$
+residual_quantization = true
+[build.internal]
+lists = [33]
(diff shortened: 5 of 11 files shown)


src/index/vchordrq/types.rs
```diff
@@ -44,6 +44,10 @@ pub struct VchordrqInternalBuildOptions {
     #[serde(default = "VchordrqInternalBuildOptions::default_build_threads")]
     #[validate(range(min = 1, max = 255))]
     pub build_threads: u16,
+    #[serde(default = "VchordrqInternalBuildOptions::default_approximate")]
+    pub approximate: bool,
+    #[serde(default = "VchordrqInternalBuildOptions::default_approximate_reduction_dim")]
+    pub approximate_reduction_dim: u32,
```
usamoi: You can use a tagged union instead of a bool.

```rust
#[derive(Deserialize)]
#[serde(untagged)]
pub enum VchordrqInternalBuildOptions {
    Lloyd(LloydOptions),
    Reduction(ReductionOptions),
}

#[derive(Deserialize)]
pub struct LloydOptions {}

#[derive(Deserialize)]
#[serde(tag = "algorithm")]
pub struct ReductionOptions {}
```

```toml
build.internal = {} # old behavior
build.internal = { algorithm = "reduction" } # new behavior
```

src/index/vchordrq/am/am_build.rs
```diff
@@ -1324,6 +1372,201 @@ fn make_default_build(
...
+        } else {
+            match self.priority.partial_cmp(&other.priority) {
+                Some(ordering) => ordering,
+                None => std::cmp::Ordering::Equal,
+            }
+        }
+    }
+}
+
+fn make_internal_approximate_build(
```
usamoi: We might prefer to move the KMeans-related code to `crates/k_means`.

crates/k_means/src/reduction.rs
```diff
@@ -0,0 +1,237 @@
...
+            .map(|s| {
+                (0..self.in_dims as usize)
+                    .map(|i| f32::reduce_sum_of_xy(&self.inv_projector[i], &s))
+                    .collect()
+            })
+            .collect()
+    }
+}
+
+fn cholesky_with_shift(m: &[Vec<f32>]) -> Vec<Vec<f32>> {
```
usamoi:
```suggestion
fn cholesky_with_shift(m: &[impl AsRef<f32>]) -> Vec<Vec<f32>> {
```

crates/k_means/src/reduction.rs
```diff
@@ -0,0 +1,237 @@
...
+            for j in (i + 1)..n {
+                cholesky_l[i][j] = 0.0;
+            }
+        }
+        if ok {
+            return cholesky_l;
+        }
+        lambda *= 10.0;
+    }
+    panic!("Cholesky failed even after diagonal shift");
```
usamoi: For a symmetric positive definite matrix, wouldn't this always succeed?

src/index/vchordrq/am/am_build.rs
cutecutecat: This is the designed changes:
- `k_means_lookup` is moved into `crates/k_means`
- `k_means_centroids_inplace` is created in `crates/k_means`, used to calculate centroids from sum/count
- `sainte_lague_allocate` is moved to `crates/k_means`, based on https://en.wikipedia.org/wiki/Sainte-Lagu%C3%AB_method

crates/k_means/src/reduction.rs
usamoi: Why it's marked resolved without replies?

crates/k_means/src/reduction.rs
```diff
@@ -0,0 +1,183 @@
...
+pub struct KMeansReduction {
+    pub projector: Vec<Vec<f32>>,
+    pub in_dims: u32,
+    pub out_dims: u32,
+}
+
+impl KMeansReduction {
+    pub fn new(in_dims: u32, out_dims: u32) -> Self {
+        assert!(in_dims >= out_dims);
+        let mut rng = StdRng::from_seed([7; 32]);
```
usamoi: > Struct StdRng
> A strong, fast (amortized), non-portable RNG

Use `rand_chacha`.

src/index/vchordrq/am/am_build.rs
```diff
@@ -1324,6 +1379,226 @@ fn make_default_build(
...
+            "approximation: restoring the precise index structure and dimension from {} samples.",
+            samples_idx.len()
+        );
+        let dim = vector_options.dims as usize;
+        let bottom = tree.first().unwrap();
+        let top = tree.last().unwrap();
+        let mut centroids = vec![vec![0.0_f32; dim]; bottom.len()];
+        let mut count = vec![0_u32; bottom.len()];
+        let mut idx_heap = 0_usize;
+        let mut idx_alloc = 0_usize;
```
usamoi: Order is not necessarily preserved across multiple scans. Use `HeapFetcher`.

crates/k_means/src/reduction.rs
cutecutecat: This is removed, now we don't need to recover dimension at all.

crates/k_means/src/reduction.rs
cutecutecat: Here is deprecated now. The reduction is doing by rotate then truncate.

src/index/vchordrq/am/am_build.rs
cutecutecat: To use `HeapFetcher`, I found it is required to: 
- init a temporary heapfetch from `index_fetch_begin` 
- and drop it by `index_fetch_end`
- init a new slot allocated by `ExecInitExtraTupleSlot`( `table_slot_create` will trigger an assert https://github.com/postgres/postgres/blob/7885b94dd81b98bbab9ed878680d156df7bf857f/src/backend/access/heap/heapam_handler.c#L123)

The new initialiazation is put inside `new_with_tmp_heapfetch`. Above is the the design concept, you can review the code based on it.

usamoi: Why the assertion is triggered? This shouldn't happen.

usamoi: `ExecInitExtraTupleSlot` is definitely incorrect. It just happens to work.

src/index/vchordrq/am/am_build.rs
```diff
@@ -1287,6 +1339,252 @@ fn make_default_build(
...
+    allocation: HashMap<usize, usize>,
+) -> Vec<Vec<f32>> {
+    use crate::index::fetcher::Tuple;
+    let mut centroids = vec![vec![0.0_f32; dim]; centroids_len];
+    let mut idx_alloc = 0_usize;
+    let mut count = vec![0_u32; centroids_len];
+    for ctid in samples_idx.into_iter() {
+        let key = ctid_to_key(ctid);
+        let Some(mut tuple) = fetcher.fetch(key) else {
+            pgrx::error!("the table format is incorrect");
```
usamoi: What does it mean?

src/index/vchordrq/am/am_build.rs
```diff
@@ -1287,6 +1339,252 @@ fn make_default_build(
...
+        internal_build.kmeans_iterations as _,
+    );
+    result.push(Structure {
+        centroids: top_centroids.clone(),
+        children: vec![(0..centroids.len() as u32).collect(); 1],
+    });
+    result
+}
+
+fn traverse_heap_again(
```
usamoi: We don't necessarily have to rely on obtaining the original vectors. We can either store the original vectors in a temporary file, or start an independent sampling process again and compute the centroids by observing the newly sampled data.

src/index/vchordrq/am/am_build.rs
```diff
@@ -232,6 +235,19 @@ pub unsafe extern "C-unwind" fn ambuild(
             };
             let sampler = unsafe { HeapSampler::new(index_relation, heap_relation, snapshot) };
             let mut sample = sampler.sample();
+            let (approximation, reduction, k_means_dim) = match &internal_build.kmeans_algorithm {
```
usamoi: Why not match `internal_build.kmeans_algorithm` directly?

src/index/vchordrq/am/am_build.rs
```diff
@@ -1287,6 +1339,252 @@ fn make_default_build(
...
+    samples: Vec<Vec<f32>>,
+    reporter: &PostgresReporter,
+) -> (Vec<Structure<Normalized>>, HashMap<usize, usize>) {
+    let mut result = Vec::<Structure<Vec<f32>>>::new();
+    let (top_list, bottom_list) = if internal_build.lists.len() == 1 {
+        let top_k = (internal_build.lists[0] as f64).sqrt().floor() as u32;
+        let top_k = top_k.clamp(1, (samples.len() as f64).sqrt().floor() as u32);
+        (top_k, internal_build.lists[0])
+    } else {
+        (internal_build.lists[0], internal_build.lists[1])
```
usamoi: `lists` should be independent.

src/index/vchordrq/am/am_build.rs
cutecutecat: Fixed.

src/index/vchordrq/am/am_build.rs
cutecutecat: OK, I think this could be `unreachable!()`

src/index/vchordrq/am/am_build.rs
cutecutecat: Ok, I checked her again, and find table_slot_create is worked now. The assert might be triggered by elsewhere.

src/index/vchordrq/am/am_build.rs
cutecutecat: Fixed with new function `approx_kmeans` which only generate centroids. The recall/QPS is still the same like before.

src/index/vchordrq/am/am_build.rs
cutecutecat: Fixed with sampling again and centroids are computed by new ones.

crates/k_means/src/lib.rs
```diff
@@ -32,7 +33,7 @@ pub struct This {
 
 pub trait KMeans {
     fn this(&mut self) -> &mut This;
-    fn assign(&mut self);
+    fn assign(&mut self) -> Vec<usize>;
```
usamoi: This should not return any value. If necessary, it should be recalculated.

crates/k_means/src/lib.rs
```diff
@@ -46,6 +47,7 @@ pub trait KMeans {
         });
     }
     fn finish(self: Box<Self>) -> Square;
+    fn finish_samples(self: Box<Self>) -> (Square, Square);
```
usamoi: This should not return samples. The implementation details of K-means can certainly involve some irreversible operations.

src/index/vchordrq/types.rs
```diff
@@ -44,6 +53,8 @@ pub struct VchordrqInternalBuildOptions {
     #[serde(default = "VchordrqInternalBuildOptions::default_build_threads")]
     #[validate(range(min = 1, max = 255))]
     pub build_threads: u16,
+    #[serde(default = "VchordrqInternalBuildOptions::default_kmeans_algorithm")]
+    pub kmeans_algorithm: KMeansAlgorithm,
```
usamoi, line 56: Dimension reduction can be combined with any specific K-means algorithm. It is unreasonable to couple dimension reduction with performing K-means on the partitions.

src/index/vchordrq/am/am_build.rs
```diff
@@ -234,8 +235,25 @@ pub unsafe extern "C-unwind" fn ambuild(
...
+                    (true, false, vector_options.dims as usize)
+                }
+                KMeansAlgorithm::Reduction { dim: Some(d) } => (true, true, d as usize),
+            };
+            assert!(
+                k_means_dim <= vector_options.dims as usize,
+                "invalid K-means reduction dimension {} > {}",
+                k_means_dim,
+                vector_options.dims
+            );
```
usamoi: It should be checked eariler.

src/index/vchordrq/am/am_build.rs
```diff
@@ -234,8 +235,25 @@ pub unsafe extern "C-unwind" fn ambuild(
...
             let sampler = unsafe { HeapSampler::new(index_relation, heap_relation, snapshot) };
             let mut sample = sampler.sample();
+            let (approximation, reduction, k_means_dim) = match internal_build.kmeans_algorithm {
+                KMeansAlgorithm::Lloyd {} => (false, false, vector_options.dims as usize),
+                KMeansAlgorithm::Reduction { dim: None } => {
+                    (true, false, vector_options.dims as usize)
+                }
+                KMeansAlgorithm::Reduction { dim: Some(d) } if d == vector_options.dims => {
+                    (true, false, vector_options.dims as usize)
+                }
```
usamoi: Why do we need this special condition?

src/index/vchordrq/am/am_build.rs
```diff
@@ -234,8 +235,25 @@ pub unsafe extern "C-unwind" fn ambuild(
...
+                KMeansAlgorithm::Reduction { dim: Some(d) } => (true, true, d as usize),
+            };
+            assert!(
+                k_means_dim <= vector_options.dims as usize,
+                "invalid K-means reduction dimension {} > {}",
+                k_means_dim,
+                vector_options.dims
+            );
+            pgrx::info!("start collecting samples from the table");
```
usamoi:
```suggestion
```

src/index/vchordrq/am/am_build.rs
```diff
@@ -275,18 +296,34 @@ pub unsafe extern "C-unwind" fn ambuild(
...
+                        k_means_dim,
+                        internal_build.clone(),
+                        samples,
+                        &reporter,
+                        opfamily,
+                        &mut sample,
+                    )
+                }
+                false => make_internal_build(vector_options, internal_build, samples, &reporter),
+            };
```
usamoi: There is no reason to have two separate paths here.

src/index/vchordrq/am/am_build.rs
```diff
@@ -275,18 +296,34 @@ pub unsafe extern "C-unwind" fn ambuild(
...
             }
-            make_internal_build(vector_options, internal_build, samples, &reporter)
+            tree
         }
         VchordrqBuildSourceOptions::External(external_build) => {
             reporter.phase(BuildPhase::from_code(BuildPhaseCode::ExternalBuild));
             make_external_build(vector_options, opfamily, external_build)
         }
     };
+    pgrx::info!("the K-means structure tree is constructed");
```
usamoi:
```suggestion
```

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,253 @@
...
+
+pub struct ApproxKMeans {
+    pub d: usize,
+    pub top_centroids: Square,
+    pub bottom_centroids: Square,
+    pub bottom_partition: Vec<Square>,
+    pub partition_start: Vec<usize>,
+}
+
+pub fn approx_kmeans(
```
usamoi: Why doesn't it follow what other algorithms do?

crates/k_means/src/lib.rs
cutecutecat: Now we recalculate by a new function `k_means_assign`.

crates/k_means/src/lib.rs
cutecutecat: Fixed.

src/index/vchordrq/types.rs
cutecutecat: Now the reduction dimension can be used with both `lloyd` or `approx`, and is being seperated from algorithm. The syntax is like this, I think it could be refined later.
```sql
CREATE INDEX ON t USING vchordrq (val vector_l2_ops)
WITH (options = $$
residual_quantization = false
[build.internal]
lists = [3]
spherical_centroids = false
kmeans_dimension_reduction = 2
[build.internal.kmeans_algorithm.approx]
```

src/index/vchordrq/am/am_build.rs
cutecutecat: Removed.

src/index/vchordrq/am/am_build.rs
cutecutecat: Merged into `make_internal_build` with internal kmeans choice.

crates/k_means/src/approx.rs
cutecutecat: Now the approx is implemented with train KMeans.

src/index/vchordrq/types.rs
```diff
@@ -27,6 +27,15 @@ impl Default for VchordrqDefaultBuildOptions {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub enum KMeansAlgorithm {
+    #[serde(rename = "lloyd")]
+    Lloyd {},
+    #[serde(rename = "approx")]
+    Approx {},
+}
```
usamoi, line 36:
```suggestion
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(deny_unknown_fields)]
#[serde(rename_all = "snake_case")]
pub enum KMeansAlgorithm {
    Lloyd {},
    Approximate {},
}
```

src/index/vchordrq/am/am_build.rs
```diff
@@ -1274,8 +1276,17 @@ fn make_internal_build(
...
     use humansize::{BINARY, format_size};
     use std::iter::once;
+    let (reduction, sample_dim) = match internal_build.kmeans_dimension_reduction {
+        None => (false, vector_options.dims as usize),
+        Some(d) if d < vector_options.dims => (true, d as usize),
+        Some(d) => pgrx::error!(
+            "invalid K-means reduction dimension {} >= {}",
+            d,
+            vector_options.dims
+        ),
```
usamoi: Here, some people might prefer to make it only a warning and revert to the default behavior. In this way, by setting this value, they can ensure an upper limit on the time consumption, regardless of the dimension of columns.

src/index/vchordrq/am/am_build.rs
```diff
@@ -1316,9 +1326,12 @@ fn make_internal_build(
                             x.len() as u32,
                             "invalid vector dimensions"
                         );
+                        if reduction {
+                            x = rabitq::rotate::rotate(&x)[..sample_dim].to_vec();
```
usamoi:
```suggestion
                            rabitq::rotate::rotate_inplace(&mut x);
                            x.truncate(simple_dim as _);
```

src/index/vchordrq/am/am_build.rs
```diff
@@ -1327,9 +1340,16 @@ fn make_internal_build(
         samples
     };
     let mut result = Vec::<Structure<Normalized>>::new();
+    let approximation = match internal_build.kmeans_algorithm {
+        KMeansAlgorithm::Lloyd {} => false,
+        KMeansAlgorithm::Approx {} => true,
+    };
+    if !approximation && reduction {
+        pgrx::warning!("using dimension reduction without approximation could be much slower.");
```
usamoi: I guess that it means here is that the acceleration is not maximized. I'm not sure whether it's useful for us to set this warning. Maybe, some people use dimension reduction simply to reduce memory usage.

src/index/vchordrq/am/am_build.rs
```diff
@@ -1346,6 +1366,7 @@ fn make_internal_build(
         let num_dims = vector_options.dims as usize;
         let num_lists = w as usize;
         let num_iterations = internal_build.kmeans_iterations as _;
+        let input_dim = input.d();
```
usamoi:
```suggestion
       let num_dims = input.d();
       let num_lists = w as usize;
       let num_iterations = internal_build.kmeans_iterations as _;
```

This helps the log in the next block to display the correct dimension.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,299 @@
...
+    pool.install(|| {
+        targets
+            .par_iter_mut()
+            .zip(samples.par_iter_mut())
+            .for_each(|(target, sample)| {
+                *target = crate::flat::k_means_lookup(sample, centroids);
+            });
+    });
+    targets
+}
```
usamoi: This function is unnecessary, and it strangely creates the thread pool twice.

By the way, we should stop passing in `num_threads` and pass in `&ThreadPool` instead. Otherwise, this will create unnecessary threads. On my system, the maximum number of threads a process can create is 248,567 (`cat /proc/sys/kernel/threads-max`).

crates/k_means/src/square.rs
```diff
@@ -114,3 +121,168 @@ impl<'a> rayon::prelude::IntoParallelIterator for &'a mut Square {
...
+            if i != j {
+                let start_i = i * d;
+                let start_j = j * d;
+                for offset in 0..d {
+                    data.swap(start_i + offset, start_j + offset);
+                }
+            }
+            to.swap(i, j);
+        }
+    }
```
usamoi: Use https://doc.rust-lang.org/std/primitive.slice.html#method.swap_with_slice.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,299 @@
...
+            targets,
+            rng,
+        },
+        top_centroids,
+        partition_start,
+        bottom_k_means,
+    })
+}
+
+pub fn k_means_dimension_restore<'a>(
```
usamoi: This function shouldn’t be here either.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+fn k_means_centroids_inplace(centroids: &mut [f32], count: u32, is_spherical: bool) {
+    assert!(!centroids.is_empty());
+    assert!(count > 0);
+    let dim = centroids.len();
+    let c = 1.0 / count as f32;
+    for d in 0..dim {
+        centroids[d] *= c;
+    }
```
usamoi: Use `simd::Floating::vector_mul_scalar_inplace`.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,299 @@
...
+
+pub fn k_means_dimension_restore<'a>(
+    d: usize,
+    c: usize,
+    num_points: usize,
+    is_spherical: bool,
+    index: Box<dyn Fn(&[f32]) -> u32 + 'a>,
+    mut next_sample: impl FnMut() -> Option<(Vec<f32>, Vec<f32>)>,
+) -> Square {
+    let mut centroids = vec![vec![0.0_f32; d]; c];
```
usamoi: Use `Square`.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,299 @@
...
+    let mut ret = Square::new(d);
+    let mut count = vec![0_u32; c];
+    let mut traveled = 0;
+    while let Some((original, reduction)) = next_sample()
+        && traveled < num_points
+    {
+        let cid = index(&reduction) as usize;
+        centroids[cid] = f32::vector_add(&centroids[cid], &original);
```
usamoi: There should be `simd::Floating::vector_add_scalar_inplace`. You could add that.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,299 @@
...
+        count[cid] += 1;
+        traveled += 1;
+    }
+    (centroids, count) = centroids
+        .into_iter()
+        .zip(count)
+        .filter(|(_, cnt)| *cnt > 0)
+        .unzip();
+    for i in 0..count.len() {
```
usamoi: We'd better ensure that exactly `c` centroids are returned. The other KMeans functions all guarantee this.

src/index/vchordrq/types.rs
```diff
@@ -44,6 +52,10 @@ pub struct VchordrqInternalBuildOptions {
     #[serde(default = "VchordrqInternalBuildOptions::default_build_threads")]
     #[validate(range(min = 1, max = 255))]
     pub build_threads: u16,
+    #[serde(default = "VchordrqInternalBuildOptions::default_kmeans_algorithm")]
+    pub kmeans_algorithm: KMeansAlgorithm,
+    #[serde(default = "VchordrqInternalBuildOptions::default_kmeans_dimension_reduction")]
+    pub kmeans_dimension_reduction: Option<u32>,
```
usamoi, line 59:
```suggestion
    #[serde(default = "VchordrqInternalBuildOptions::default_kmeans_dimension_reduction")]
    #[validate(range(min = 1, max = 16000))]
    pub kmeans_dimension_reduction: Option<u32>,
```

tests/vchordrq/approximate.slt
```diff
@@ -0,0 +1,57 @@
...
+
+statement ok
+CREATE INDEX ON t USING vchordrq (val vector_l2_ops)
+WITH (options = $$
+residual_quantization = true
+[build.internal]
+lists = [33]
+spherical_centroids = false
+kmeans_dimension_reduction = 2
+kmeans_algorithm = { hierarchical = {} }
```
usamoi: nit

```suggestion
kmeans_algorithm.hierarchical = {}
```

crates/k_means/src/square.rs
```diff
@@ -114,3 +121,174 @@ impl<'a> rayon::prelude::IntoParallelIterator for &'a mut Square {
...
+}
+
+pub fn partition_mut<'a>(sq: SquareMut<'a>, alloc: &[Vec<usize>]) -> Vec<SquareMut<'a>> {
+    let d = sq.d;
+    let data = sq.p;
+    let n = data.len() / d;
+
+    let total: usize = alloc.iter().map(|g| g.len()).sum();
+    assert_eq!(total, n, "alloc must cover all rows without duplication");
```
usamoi:
```suggestion
```

It's just `order.len()`, which is checked later.

crates/k_means/src/square.rs
```diff
@@ -114,3 +121,174 @@ impl<'a> rayon::prelude::IntoParallelIterator for &'a mut Square {
...
+    let mut slice: &mut [f32] = data;
+    for g in alloc {
+        let elems = g.len() * d;
+        let (head, rest) = std::mem::take(&mut slice).split_at_mut(elems);
+        out.push(SquareMut { d, p: head });
+        slice = rest;
+    }
+    debug_assert!(slice.is_empty());
+    out
+}
```
usamoi:
```suggestion
pub fn partition_mut<'a>(
    mut a: SquareMut<'a>,
    groups: &[impl AsRef<[usize]>],
) -> Vec<SquareMut<'a>> {
    let n = a.len();

    let permutation = groups
        .iter()
        .flat_map(AsRef::as_ref)
        .copied()
        .collect::<Vec<_>>();

    let mut marked = vec![false; a.len()];
    let mut buffer = vec![0.0; a.d()];

    for i in 0..n {
        if marked[i] {
            continue;
        }
        buffer.copy_from_slice(&a[i]);
        let (mut src, mut dst) = (permutation[i], i);
        while src != i {
            a.copy_within(src..src + 1, dst);
            marked[dst] = true;
            (src, dst) = (permutation[src], src);
        }
        a[dst].copy_from_slice(&buffer);
        marked[dst] = true;
    }

    let mut result = Vec::with_capacity(groups.len());

    let (d, mut p) = a.into_inner();

    for group in groups {
        let group = group.as_ref();
        let head;
        (head, p) = std::mem::take(&mut p).split_at_mut(group.len() * d);
        result.push(SquareMut::new(d, head));
    }

    result
}
```

src/index/vchordrq/am/am_build.rs
```diff
@@ -1327,9 +1343,13 @@ fn make_internal_build(
         samples
     };
     let mut result = Vec::<Structure<Normalized>>::new();
+    let approximation = match internal_build.kmeans_algorithm {
+        KMeansAlgorithm::Lloyd {} => false,
+        KMeansAlgorithm::Hierarchical {} => true,
+    };
```
usamoi: Just use `match` where `approximation` is used.

src/index/vchordrq/am/am_build.rs
```diff
@@ -1274,8 +1275,20 @@ fn make_internal_build(
 ) -> Vec<Structure<Normalized>> {
     use humansize::{BINARY, format_size};
     use std::iter::once;
+    let (reduction, sample_dim) = match internal_build.kmeans_dimension_reduction {
+        None => (false, vector_options.dims as usize),
+        Some(d) if d < vector_options.dims => (true, d as usize),
+        Some(d) => {
+            pgrx::warning!(
+                "invalid K-means reduction dimension {} >= {}, reduction is not enabled",
```
usamoi: The wording here is not very good.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,238 @@
...
+        } else if other.priority.is_nan() {
+            std::cmp::Ordering::Greater
+        } else {
+            match self.priority.partial_cmp(&other.priority) {
+                Some(ordering) => ordering,
+                None => std::cmp::Ordering::Equal,
+            }
+        }
+    }
+}
```
usamoi: This is not a correct implementation of partial ordering. If the standard library algorithms change, it could break your code.

[https://doc.rust-lang.org/std/cmp/trait.Ord.html](https://doc.rust-lang.org/std/cmp/trait.Ord.html)

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,238 @@
...
+    let mut alloc_lists = vec![1u32; proportion.len()];
+    let mut diff = all_clusters as i32 - proportion.len() as i32;
+    if diff < 0 {
+        panic!(
+            "build.lists is too large: requested {}, but only {} are available.",
+            all_clusters,
+            proportion.len()
+        );
+    }
+    let mut priorities: BinaryHeap<PriorityItem> = proportion
```
usamoi: My advice is just to use `(AlwaysEqual<usize>, i64)` here. You could convert `f64` to `i64`. See also `distance::Distance`. If you don't care about precision, using `Distance` is okay. If you care, use `ordered_float`.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,238 @@
...
+fn successive_quotients_allocate(all_clusters: usize, proportion: Vec<u32>) -> Vec<u32> {
+    let mut alloc_lists = vec![1u32; proportion.len()];
+    let mut diff = all_clusters as i32 - proportion.len() as i32;
+    if diff < 0 {
+        panic!(
+            "build.lists is too large: requested {}, but only {} are available.",
+            all_clusters,
+            proportion.len()
+        );
+    }
```
usamoi: We'd better handle this in a more robust way. Unlike configuration, users don't want to do data cleaning themselves.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,238 @@
...
+            index: i,
+            priority: *x as f64 / (alloc_lists[0] as f64),
+        })
+        .collect();
+    while diff > 0 {
+        let top = priorities.pop().unwrap();
+        alloc_lists[top.index] += 1;
+        priorities.push(PriorityItem {
+            index: top.index,
+            priority: proportion[top.index] as f64 / (alloc_lists[top.index] as f64),
```
usamoi:
```suggestion
            priority: proportion[top.index] as f64 / (alloc_lists[top.index] as f64 + 0.5),
```

According to [Wikipedia](https://en.wikipedia.org/wiki/Sainte-Lagu%C3%AB_method).

src/index/vchordrq/am/am_build.rs
cutecutecat: I refined it with codex, you can comment again if you think its still not good enough.

```rust
pgrx::warning!(
                "ignoring kmeans_dimension_reduction={} because it must be less than vector dimension {}; disabling reduction",
                d,
                vector_options.dims
            );
```

src/index/vchordrq/am/am_build.rs
```diff
@@ -1428,6 +1491,46 @@ fn make_internal_build(
...
+        let cid = index(&reduction) as usize;
+        f32::vector_add_inplace(&mut centroids[cid], &original);
+        count[cid] += 1;
+        traveled += 1;
+    }
+    let mut ret = Square::new(d);
+    for (i, centroid) in centroids.as_mut_view().iter_mut().enumerate() {
+        if count[i] == 0 {
+            let mut c: Vec<f32> = (0..d).map(|_| rng.random_range(-1.0f32..1.0f32)).collect();
+            centroid.swap_with_slice(c.as_mut_slice());
```
usamoi:
```suggestion
            centroid.fill_with(|| rng.random_range(-1.0f32..1.0f32));
```

src/index/vchordrq/am/am_build.rs
```diff
@@ -1327,9 +1343,13 @@ fn make_internal_build(
         samples
     };
     let mut result = Vec::<Structure<Normalized>>::new();
+    // let approximation = match internal_build.kmeans_algorithm {
+    //     KMeansAlgorithm::Lloyd {} => false,
+    //     KMeansAlgorithm::Hierarchical {} => true,
+    // };
```
usamoi:
```suggestion
```

crates/k_means/src/square.rs
```diff
@@ -114,3 +121,183 @@ impl<'a> rayon::prelude::IntoParallelIterator for &'a mut Square {
...
+    type Item = &'a mut [f32];
+
+    type Iter = rayon::slice::ChunksExactMut<'a, f32>;
+
+    fn into_par_iter(self) -> Self::Iter {
+        rayon::slice::ParallelSliceMut::par_chunks_exact_mut(self.p, self.d)
+    }
+}
+
+pub fn partition_mut<'a>(
```
usamoi: Move it to `approx.rs`.

crates/k_means/src/approx.rs
```diff
@@ -0,0 +1,207 @@
...
+use crate::square::{Square, SquareMut, partition_mut};
+use crate::{KMeans, This};
+use always_equal::AlwaysEqual;
+use distance::Distance;
+use rand::SeedableRng;
+use rand::rngs::StdRng;
+use rayon::iter::{IndexedParallelIterator, IntoParallelRefMutIterator, ParallelIterator};
+use std::collections::BinaryHeap;
+
+struct Approx<'a> {
```
usamoi: Please change all file names and variable names from `approx` to `hierarchical`.

src/index/vchordrq/am/am_build.rs
```diff
@@ -1274,8 +1275,20 @@ fn make_internal_build(
 ) -> Vec<Structure<Normalized>> {
     use humansize::{BINARY, format_size};
     use std::iter::once;
+    let (reduction, sample_dim) = match internal_build.kmeans_dimension_reduction {
+        None => (false, vector_options.dims as usize),
+        Some(d) if d < vector_options.dims => (true, d as usize),
+        Some(d) => {
+            pgrx::warning!(
+                "ignoring kmeans_dimension_reduction={} because it must be less than vector dimension {}; disabling reduction",
```
usamoi:
```suggestion
                "ignoring `kmeans_dimension_reduction = {}` because it is less than the vector dimension {}",
```

usamoi (approved):

cutecutecat:
Blocking: I will merge it when the final precision test is finished.

PR #403, opened 2026-01-08
chore: update dependencies

reneleonhardt:
### Chores
- Update cargo dependencies, CI dependencies and GitHub Actions (using pinning to improve supply chain security)
- Let Dependabot update github-actions and cargo (respecting a cooldown to improve supply chain security)
- Dropped PostgreSQL 13 from checks, release and runtime requirements (EOL since 2025-11-13)
- Bumped Debian 12 (Bookworm) to 13 (Trixie), and therefore GCC 12 to 14
- Bumped clang 18 to 21

### DIFF
--- a/src/index/vchordg/am/am_build.rs
+++ b/src/index/vchordg/am/am_build.rs
@@ -560,12 +560,7 @@ unsafe fn options(
     index_relation: pgrx::pg_sys::Relation,
 ) -> (VectorOptions, VchordgIndexingOptions) {
     let att = unsafe { &mut *(*index_relation).rd_att };
-    #[cfg(any(
-        feature = "pg14",
-        feature = "pg15",
-        feature = "pg16",
-        feature = "pg17"
-    ))]
+    #[cfg(any(feature = "pg14", feature = "pg15", feature = "pg16", feature = "pg17"))]
     let atts = unsafe { att.attrs.as_slice(att.natts as _) };
     #[cfg(feature = "pg18")]
     let atts = unsafe {
--- a/src/index/vchordg/am/mod.rs
+++ b/src/index/vchordg/am/mod.rs
@@ -241,12 +241,7 @@ pub unsafe extern "C-unwind" fn ambulkdelete(
     let opfamily = unsafe { opfamily((*info).index) };
     let index = unsafe { PostgresRelation::new((*info).index) };
     let check = || unsafe {
-        #[cfg(any(
-            feature = "pg14",
-            feature = "pg15",
-            feature = "pg16",
-            feature = "pg17"
-        ))]
+        #[cfg(any(feature = "pg14", feature = "pg15", feature = "pg16", feature = "pg17"))]
         pgrx::pg_sys::vacuum_delay_point();
         #[cfg(feature = "pg18")]
         pgrx::pg_sys::vacuum_delay_point(false);
@@ -278,12 +273,7 @@ pub unsafe extern "C-unwind" fn amvacuumcleanup(
     let opfamily = unsafe { opfamily((*info).index) };
     let index = unsafe { PostgresRelation::new((*info).index) };
     let check = || unsafe {
-        #[cfg(any(
-            feature = "pg14",
-            feature = "pg15",
-            feature = "pg16",
-            feature = "pg17"
-        ))]
+        #[cfg(any(feature = "pg14", feature = "pg15", feature = "pg16", feature = "pg17"))]
         pgrx::pg_sys::vacuum_delay_point();
         #[cfg(feature = "pg18")]
         pgrx::pg_sys::vacuum_delay_point(false);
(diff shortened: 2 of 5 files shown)


reneleonhardt:

.github/workflows/release.yml
```diff
@@ -39,9 +39,9 @@ jobs:
     needs: ["semver"]
     strategy:
       matrix:
-        version: ["13", "14", "15", "16", "17", "18"]
+        version: ["14", "15", "16", "17", "18"]
         arch: ["x86_64", "aarch64"]
-    runs-on: ${{ matrix.arch == 'x86_64' && 'ubuntu-22.04' || 'ubuntu-22.04-arm' }}
+    runs-on: ${{ matrix.arch == 'x86_64' && 'ubuntu-24.04' || 'ubuntu-24.04-arm' }}
```
usamoi: This raises the minimum required libc version for the binary package, which is unacceptable.

.github/workflows/check.yml
```diff
@@ -126,11 +126,11 @@ jobs:
 
       - name: Install Clang
         run: |
-          curl --proto '=https' --tlsv1.2 -sSf https://apt.llvm.org/llvm.sh | sudo bash -s -- 18
-          echo CC=clang-18 >> $GITHUB_ENV
+          curl --proto '=https' --tlsv1.2 -sSf https://apt.llvm.org/llvm.sh | sudo bash -s -- 21
```
usamoi: We would like to support older versions of Clang, since using the latest C compiler on Linux is not always easy, and upgrading Clang here would make it difficult for us to ensure compatibility between the code and the compiler.

.github/workflows/check.yml
```diff
@@ -351,16 +351,16 @@ jobs:
...
 
-          curl -fsSL https://github.com/risinglightdb/sqllogictest-rs/releases/download/v0.28.4/sqllogictest-bin-v0.28.4-${{ matrix.arch }}-apple-darwin.tar.gz | tar -xOzf - ./sqllogictest | tee /usr/local/bin/sqllogictest > /dev/null && chmod 755 /usr/local/bin/sqllogictest
+          curl -fsSL https://github.com/risinglightdb/sqllogictest-rs/releases/download/v0.29.0/sqllogictest-bin-v0.29.0-${{ matrix.arch }}-apple-darwin.tar.gz | tar -xOzf - ./sqllogictest | tee /usr/local/bin/sqllogictest > /dev/null && chmod 755 /usr/local/bin/sqllogictest
 
       - name: Install Rust
         run: rustup update
 
       - name: Install Clang
-        run: echo CC=$(brew --prefix llvm@18)/bin/clang >> $GITHUB_ENV
+        run: echo CC=$(brew --prefix llvm@21)/bin/clang >> $GITHUB_ENV
```
usamoi: Only llvm 18 is preinstalled in the image.

https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md

.github/workflows/check.yml
```diff
@@ -687,50 +684,44 @@ jobs:
             platform: "s390x"
             clang_triple: "s390x-linux-gnu"
             rust_triple: "s390x-unknown-linux-gnu"
-            gcc_version: "12"
-            debian_version: "bookworm"
+            gcc_version: "14"
+            debian_version: "trixie"
```
usamoi: PostgreSQL does not officially maintain repositories for s390x and riscv64. Therefore, the PostgreSQL debian packages here are downloaded from the official Debian repositories. That is to say, `bookworm` can only be used to test PostgreSQL 15, and PostgreSQL 15 can only be tested on `bookworm`. PostgreSQL 15 can't be tested on `trixie`.

Cargo.toml
usamoi: To drop support for PostgreSQL 13, we would need to remove this line, and remove the parts of the code that are conditionally compiled with `feature = "pg13"`.

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
+    cooldown: # applies only to non-security updates
+      semver-patch: 3 # wait 3 days before applying patch updates
+      semver-minor: 7
+      semver-major: 14
```
usamoi: Is it possible to limit it to security updates only?

Regular updates frequently introduce breaking changes and duplicate versions of the same dependency in the dependency tree, which is a cause for concern. We would rather manage those updates manually. We might prefer to use [`pubtime`](https://blog.rust-lang.org/inside-rust/2026/01/07/this-development-cycle-in-cargo-1.93/#pubtime) locally to implement the cooldown for the new version.

usamoi (changes requested):
Please restrict the scope of this PR to updating the versions of `cargo-deny` and `sqllogictest`, and configuring Dependabot for GitHub Actions.

.github/workflows/release.yml
reneleonhardt: What about check:psql?

usamoi: It's not very important, but it would be better if both use the same version. This way, we won't run into unexpected issues during the release.

usamoi (changes requested):
Please fix the compilation warnings.

https://github.com/tensorchord/VectorChord/actions/runs/20855548368/job/59923453739?pr=403

PR #434, opened 2026-02-26
perf: disable SAQ for rabitq8

usamoi:


VoVAllen:
Add a new section in the docs to describe the different behavior comparing to fp32/fp16 vectors:
- text conversion
- no explicit cast

PR #446, opened 2026-03-23
fix: skip recorder and sqlite state creation when vchord extension is not created

Ngalstyan4:
When vchord is added to shared_preload_libraries without running CREATE EXTENSION, the object_access_hook fires on any DROP TABLE or DROP DATABASE and calls the recorder's get() function, which lazily creates the pgsql_tmp_vchord_sampling/ directory and SQLite files under PGDATA.

This is unnecessary and confusing when looking at contents of PGDATA and seeing files from an extension that has never been enabled

The commit skips all recorder operations when the extension has never been created (extension does not have an assigned OID)

### DIFF
--- a/src/recorder/worker.rs
+++ b/src/recorder/worker.rs
@@ -32,6 +32,12 @@ fn get<'a>() -> Option<RefMut<'a, rusqlite::Connection>> {
     if database_oid == 0 {
         return None;
     }
+    // Skip if vchord extension is not installed in this database
+    if unsafe { pgrx::pg_sys::get_extension_oid(c"vchord".as_ptr(), true) }
+        == pgrx::pg_sys::InvalidOid
+    {
+        return None;
+    }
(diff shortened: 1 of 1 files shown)


usamoi:
> I suspect many tools will have issue with current directory structure used in VectorChord.

Yes, it's problematic. We recommend not using this feature, as it may lead to issues. https://github.com/tensorchord/VectorChord/pull/448 should stop the file or directory creation if the feature is not used. We hope this issue can be resolved in this way.

VoVAllen:
Hi @Ngalstyan4 , the folder under `pgsql_tmp/` may still not comply with the postgres guidance to name it as `pgsql_tmpPPP.NNN`. And the lifecycle of this folder seems not clear to us and the purpose of this folder seems not consistent to what we used it for.  The ideal case is similar to `pg_stat_statements`, we're thinking about whether it's possible for us to put the sqlite file under the same folder like `pg_stat_tmp`

PR #461, opened 2026-05-21
fix(vchordrq): estimate filtered candidate selectivity

spenc-r:
## What changed

Fixes vchordrq cost estimation for filtered `ORDER BY distance LIMIT ...` queries.

The estimator now separates:

- filter selectivity: derived from the planner filtered row estimate (`baserel->rows` or parameterized `ppi_rows`)
- index selectivity: derived from the expected candidate count retrieved by the scan

This avoids reporting `indexSelectivity = 1.0` for heap-side filters that are not index quals, while also avoiding the opposite under-estimate of reporting only the final survivor fraction.

The candidate-processing term is also clamped by the IVF candidate budget (`node_count`).

## Why

For vchordrq indexes, filters on non-vector columns live in `baserestrictinfo`, not `indexclauses`. The previous estimator therefore often reported selectivity as 1.0, causing PostgreSQL to price expensive filter evaluation against the whole table and reject the vector index path even when it was the better plan.

## Tests

- [x] Added `tests/vchordrq/cost_estimator.slt`
- [x] Local sqllogictest run completed; adjusted case 11 for PG18 disabled-path behavior
- [x] `git diff --check origin/main...HEAD`
- [ ] CI

### DIFF
--- a/src/index/vchordrq/am/mod.rs
+++ b/src/index/vchordrq/am/mod.rs
@@ -292,20 +292,31 @@ pub unsafe extern "C-unwind" fn amcostestimate(
             *index_pages = 1.0;
             return;
         }
-        let selectivity = {
-            use pgrx::pg_sys::{
-                JoinType, add_predicate_to_index_quals, clauselist_selectivity,
-                get_quals_from_indexclauses,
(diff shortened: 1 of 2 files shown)


spenc-r:

PR #467, opened 2026-06-30
build: fix target features for rustc 1.96

CertainLach:
Currently, build of vchord fails on rustc 1.96 with `rustc-LLVM ERROR: Cannot select: intrinsic %llvm.x86.avx512.vpdpbusd.512`

This intrinsic is a part of `avx512f`, which doesn't seem to be enabled here. The way it worked previously is I guess because it was most likely implied by other `avx512` features

I haven't found the exact rustc bug, but it seems this kind of problems is not something new, and better avoided by specifying features explicitly

- https://github.com/rust-lang/rust/issues/134792
- https://github.com/rust-lang/rust/issues/116516
- https://github.com/rust-lang/rust/issues/125492
- https://github.com/rust-lang/rust/pull/121088


CertainLach:

CertainLach:
It seems that it was not caused by rustc this time, but incorrect rustc packaging in nixos

This patch can be merged for correctness WRT implied features (`avx512f` is a part of `x86_64-v4` after all), but should not be actually a problem

PR #468, opened 2026-06-30
fix(vchordg): avoid panic when validating an empty alpha option

ameyypawar:
## What

`VchordgIndexOptions::validate_alpha` indexes `alpha[0]` after the `is_sorted()` and range checks — both of which are vacuously true for an empty slice:

```rust
fn validate_alpha(alpha: &[f32]) -> Result<(), ValidationError> {
    if !alpha.is_sorted() { ... }                            // [].is_sorted() == true  -> skip
    if !alpha.iter().all(|x| (1.0..2.0).contains(x)) { ... } // [].iter().all(..) == true -> skip
    if alpha[0] != 1.0 { ... }                               // [] -> index out of bounds
    Ok(())
}
```

So an empty `alpha` panics (index out of bounds) instead of returning a `ValidationError`. It's reachable from SQL via `WITH (options = '[index] alpha = []')`, which surfaces a cryptic panic message rather than a clean validation error.

## Fix

Use `alpha.first()` so the empty case yields the existing `` `alpha` should contain `1.0` `` error:

```rust
if alpha.first() != Some(&1.0) {
    return Err(ValidationError::new("`alpha` should contain `1.0`"));
}
```

For any non-empty input `alpha.first() == Some(&alpha[0])`, so behavior is unchanged; only the empty slice goes from panic to a proper validation error.

## Tests

Adds unit tests for `validate_alpha` covering the empty slice (regression) plus valid/invalid cases. These run under the existing `cargo test --workspace --exclude vchord` job (`vchordg` is a pure-Rust crate).

Happy to adjust the wording or drop the tests if you'd prefer a more minimal change.


VoVAllen:
Are you using vchordg? We don't recommend it generally

ameyypawar:
No, I'm not a vchordg user — I came across this reading the option-validation code. It's a small hardening rather than something I need: `validate_alpha` indexes `alpha[0]` after the `is_sorted()` and range checks, both of which pass on an empty slice, so `alpha = []` panics instead of returning a clean validation error. The one-line `alpha.first()` change just makes that path consistent with the other checks. Completely understand if vchordg fixes aren't a priority — happy to leave the call to you.

PR #471, opened 2026-07-09
fix(gucs): report an error instead of panicking on an empty probes entry

ameyypawar:
## What

The `vchordrq.probes` parser panics on an empty entry. In `vchordrq_probes`'s `parse`:

```rust
match c {
    b' ' => continue,
    b',' => result.push(current.take().expect("empty probes")),   // panics on ",5" / "1,,2"
    b'0'..=b'9' => { /* accumulate */ }
    c => pgrx::error!("unknown character in probes: ASCII = {c}"), // clean error, same fn
}
```

A `probes` string with an empty entry — a leading comma (`',5'`) or consecutive commas (`'1,,2'`) — makes `current.take()` return `None`, so `.expect("empty probes")` **panics** instead of reporting a clean error like the sibling `unknown character` arm does. The value comes from the user-settable `vchordrq.probes` GUC (or an index `probes` reloption), so it's reachable, e.g.:

```sql
SET vchordrq.probes = ',5';
-- then a vchordrq index scan panics rather than raising a normal ERROR
```

## Fix

Handle the empty entry the same way the invalid-character case is already handled — a proper error:

```rust
b',' => match current.take() {
    Some(value) => result.push(value),
    None => pgrx::error!("empty entry in probes"),
},
```

Output is unchanged for every valid `probes` string; only the previously-panicking empty-entry case changes, now raising `ERROR: empty entry in probes`.

(No test added — `parse` lives in the root `vchord` crate, which the CI `cargo test` job excludes; the change mirrors the existing `pgrx::error!` arm.)


ameyypawar:

PR #477, opened 2026-08-23
feat(vchordrq): estimate MaxSim planner cost

HuXinjing:
## Summary

- persist the indexed vector/token count in unused `MetaTuple` padding without changing tuple offsets
- update the count after index build and bulk deletion
- add a MaxSim-specific planner model based on heap rows, indexed tokens, probes, query-token count, filters, and LIMIT
- preserve the existing planner path for every non-MaxSim opfamily

This is the first focused implementation PR proposed in #476. It does not add
an exact reranker, external tensor storage, a network protocol, or GPU code.

## Compatibility

The statistic uses the existing 2-byte and 6-byte padding regions in the meta
tuple. Compile-time offset assertions and unit tests protect the on-disk layout.

- Old indexes expose the statistic as `None`.
- The planner falls back to a configurable average document-token count.
- Build and subsequent vacuum/bulk-delete operations publish an exact count.
- No REINDEX is required solely for compatibility.

The planner query-token and fallback document-token values are configurable
until expression/page-level statistics are available.

## Validation

- `cargo test -p vchordrq`
- `cargo test --locked --workspace --exclude vchord --no-fail-fast`
- PostgreSQL 16 release extension build through the repository `xtask` build
  inside a clean PostgreSQL development container
- `cargo fmt --all -- --check`
- `git diff --check`

## Review notes

The cost constants are intentionally conservative and hardware-independent.
The model prices the eager token search and page aggregation currently
performed by MaxSim. A future optional exact-rerank backend can add its own cost
without changing this candidate-generation model.

Discussion: #476

PR #478, opened 2026-08-23
feat(vchordrq): add pluggable exact MaxSim reranking

HuXinjing:
## Summary

This is the second implementation slice proposed in Discussion #476.

It introduces a storage-independent exact MaxSim reranking boundary and a CPU reference backend. The reference tensor source reads the indexed expression from the PostgreSQL heap under the active scan snapshot; it does not introduce an external registry, network transport, GPU dependency, or quantized format.

## User-visible behavior

- Existing behavior remains the default through `vchordrq.maxsim_backend = 'coarse_only'`.
- Exact reranking is opt-in with `vchordrq.maxsim_backend = 'cpu_exact'`.
- Exact mode requires an explicit positive `vchordrq.maxsim_candidate_limit`, bounded at 65,536.
- Candidate generation remains the existing vchordrq MaxSim path. Only the bounded candidate set is rescored using the full heap tensor.

## Architecture boundary

- `CandidateTensorSource` separates full-tensor retrieval from scoring.
- `ExactMaxsimBackend` separates exact scoring from candidate generation and storage.
- `HeapTensorSource` is the only source in this PR.
- `CpuExactMaxsimBackend` is the only backend in this PR.

External tensor-source registration and any optional GPU process remain separate follow-up proposals, contingent on maintainer feedback.

## Correctness and safety

- Uses the existing `HeapFetcher`, active PostgreSQL snapshot, and indexed-expression evaluation.
- Rechecks the active scan qual before materializing a candidate tensor.
- Rejects null, mismatched-kind, and mismatched-dimension tensors.
- Does not add background workers, sockets, persistent state, migrations, or build dependencies.

## Validation

- `cargo fmt --all -- --check`
- `cargo test --locked --workspace --exclude vchord --no-fail-fast`
- PG16 release extension build and generated SQL schema
- PG16 root-extension unit test for exact result ordering
- Added `tests/vchordrq/maxsim_cpu_exact.slt`

Local validation gap: the isolated PG16 build image contains development headers but no `initdb`, so the new SQLLogicTest was not executed against a locally started PostgreSQL server. It is expected to run in the repository CI PostgreSQL matrix.

## Relationship to other work

- Discussion: #476
- Planner statistics/cost model: #477

This branch is independently based on current `main`; it does not depend on #477.

### DIFF
--- a/crates/vchordrq/src/types.rs
+++ b/crates/vchordrq/src/types.rs
@@ -18,6 +18,7 @@ use validator::{Validate, ValidationError};
 use vector::rabitq4::{Rabitq4Borrowed, Rabitq4Owned};
 use vector::rabitq8::{Rabitq8Borrowed, Rabitq8Owned};
 use vector::vect::{VectBorrowed, VectOwned};
+use vector::{VectorBorrowed, VectorOwned};
 
 #[derive(Debug, Clone, Serialize, Deserialize, Validate)]
 #[serde(deny_unknown_fields)]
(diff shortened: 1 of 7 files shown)


HuXinjing:

